PR #1, opened 2018-01-26
A first draft of a Vob.

ltratt:
The interface is designed to be a superset of `Vec<bool>`'s interface. Several functions are missing, because it would be way too much to do it all in one go. However, this contains all the functionality I need, and several things I don't, so I think it's a reasonable start.

There's extensive documentation in the commit, intentionally based on Vec's documentation whenever possible.

I'm very much looking for comments at first: this definitely isn't a "merge as soon as possible" thing!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,1231 @@
+// Copyright (c) 2018 King's College London
+// created by the Software Development Team <http://soft-dev.org/>
+//
+// The Universal Permissive License (UPL), Version 1.0
+//
+// Subject to the condition set forth below, permission is hereby granted to any person obtaining a
+// copy of this software, associated documentation and/or data (collectively the "Software"), free
+// of charge and under any and all copyright rights in the Software, and any and all patent rights
+// owned or freely licensable by each licensor hereunder covering either (i) the unmodified
+// Software as contributed to or provided by such licensor, or (ii) the Larger Works (as defined
+// below), to deal in both
+//
+// (a) the Software, and
+// (b) any piece of software and/or hardware listed in the lrgrwrks.txt file
+// if one is included with the Software (each a "Larger Work" to which the Software is contributed
+// by such licensors),
+//
+// without restriction, including without limitation the rights to copy, create derivative works
+// of, display, perform, and distribute the Software and make, use, sell, offer for sale, import,
+// export, have made, and have sold the Software and the Larger Work(s), and to sublicense the
+// foregoing rights on either these or other terms.
+//
+// This license is subject to the following condition: The above copyright notice and either this
+// complete permission notice or at a minimum a reference to the UPL must be included in all copies
+// or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+// BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+//! A vector of bits ("Vob") is a sequence of bits which exposes a `Vec`-like interface. Whereas
+//! `Vec<bool>` requires 1 byte of storage per bit, `Vob` requires only 1 bit of storage per bit.
+//!
+//! The main documentation for this crate can be found in the [`Vob`](struct.Vob.html) struct.
+
+extern crate num_traits;
+
+use std::cmp::PartialEq;
+use std::fmt;
+use std::fmt::Debug;
+use std::hash::{Hash, Hasher};
+use std::iter::FromIterator;
+use std::mem::size_of;
+use std::ops::{Index, Range};
+use std::slice;
+
+use num_traits::{PrimInt, One, Zero};
+
+
+/// A Vob is a "vector of bits": a sequence of bits which exposes a `Vec`-like interface. Whereas
+/// `Vec<bool>` requires 1 byte of storage per bit, `Vob` requires only 1 bit of storage per bit.
+/// For larger numbers of bits, Vobs can lead to a significant memory decrease and performance
+/// increase.
+///
+/// # Examples
+/// The `vob!` macro makes creating small `Vob`s easy:
+///
+/// ```rust
+/// # #[macro_use] extern crate vob;
+/// # fn main() {
+/// let mut v = vob![true, false, true];
+/// assert_eq!(v[1], false);
+/// # }
+/// ```
+///
+/// Operations such as `and`ing two `Vob`s together are quick; one can also quickly identify which
+/// bits are set:
+///
+/// ```rust
+/// use vob::Vob;
+/// let mut v1 = Vob::from_elem(256, false);
+/// let mut v2 = Vob::from_elem(256, false);
+/// v1.set(67, true);
+/// v2.set(67, true);
+/// v1.set(188, true);
+/// v1.and(&v2);
+/// let num_bits_set = v1.iter_set_bits().count();
+/// assert_eq!(num_bits_set, 1);
+/// ```
+///
+///
+/// ## Storage backing type
+///
+/// `Vob`s default to using `usize` as a storage backing type. This is generally a substantial win
+/// over using smaller storage types if you use functions such as
+/// [`or()`](struct.Vob.html#method.or). In such cases, `usize` on a 64-bit machine is almost
+/// exactly twice as fast as using `u32`. If you only ever set and get individual bits, a smaller
+/// data type might be marginally more effective: for such use cases `u32` is around 1% faster than
+/// `usize` on a 64-bit machine. You can choose your own storage type with the
+/// [`new_with_storage_type()`](struct.Vob.html#method.new_with_storage_type) constructor. In
+/// general we recommend using the default `usize` backing storage unless you have rigorously
+/// benchmarked your particular use case and are sure that a different storage type is superior.
+///
+///
+/// ## Migrating from `Vec<bool>`
+///
+/// As far as possible, `Vob` is intended to have a superset of `Vec<bool>`'s interface, which
+/// should make porting most code fairly simple. However, `Vec<bool>` contains several functions
+/// which are not yet implemented in `Vob`: these are missing simply due to a lack of a current
+/// use-case rather than because of any fundamental incompatibilities.
+///
+/// There is one missing feature which, currently, is impossible to implement: assignment to an
+/// index. In other words one cannot currently express `v[0] = true` for a `Vob` `v`. Until
+/// [`IndexGet` / `IndexMove` and equivalents](https://github.com/rust-lang/rfcs/issues/997) are
+/// implemented in `rustc`, this restriction appears to be inevitable. Note that referencing by
+/// index works (though via a hack identical to that used in `BitVec`): one can write
+/// `println!("{}", v[0])` for a `Vob` `v`, for example.
+///
+///
+/// ## Migrating from `BitVec`
+///
+/// `Vob` is directly inspired by `BitVec`, but aims to provide an interface more closely aligned
+/// to `Vec<bool>` Several functions in `BitVec` have different names in `Vob`, but porting is in
+/// general fairly simple. The main semantic difference is that `Vob`s
+/// [`clear()`](struct.Vob.html#method.clear) function empties the `Vob` of contents (i.e. sets its
+/// length to 0), whereas `BitVec`'s function of the same name unsets all bits (keeping the length
+/// unchanged). The same effect as `BitVec`'s `clear` can be achieved by using `Vob`'s
+/// [`set_all(false)`](struct.Vob.html#method.set_all) function.
+#[derive(Clone)]
+pub struct Vob<T=usize> {
+    /// How many bits are stored in this Vob?
+    len: usize,
+    /// The underlying storage. We refer to a single instance of `T` as a block. Since the storage
+    /// consists of (potentially multiple-byte) blocks, there may be "unused" bits in the final
+    /// block. We guarantee that, at all points visible to the user, the "unused" bits are set to
+    /// 0.
+    vec: Vec<T>
+}
+
+// In an ideal world, Rust's type defaults would allow us to fold the two impl blocks into one and
+// say "impl Vob<T: usize>", but currently that doesn't work.
+impl Vob<usize> {
+    /// Constructs a new, empty Vob (with `usize` backing storage, which is likely to be the best
+    /// choice in nearly all situations).
+    ///
+    /// The Vob will not allocate until elements are pushed onto it.
+    pub fn new() -> Vob<usize> {
+        Default::default()
+    }
+
+    /// Constructs a new, empty Vob (with `usize` backing storage, which is likely to be the best
+    /// choice in nearly all situations) with the specified capacity.
+    ///
+    /// The Vob will be able to hold at least `capacity` elements without reallocating. If
+    /// `capacity` is 0, the vector will not allocate.
+    pub fn with_capacity(capacity: usize) -> Vob<usize> {
+        Vob {
+            len: 0,
+            vec: Vec::with_capacity(blocks_required::<usize>(capacity))
+        }
+    }
+
+    /// Creates a BitVec that holds `len` elements, setting each element to `value`.
+    ///
+    /// # Examples
+    /// ```
+    /// use vob::Vob;
+    /// let v = Vob::from_elem(2, true);
+    /// assert_eq!(v.len(), 2);
+    /// assert_eq!(v.get(0), Some(true));
+    /// ```
+    pub fn from_elem(len: usize, value: bool) -> Vob<usize> {
+        let mut v = Vob::with_capacity(len);
+        for _ in 0..blocks_required::<usize>(len) {
+            v.vec.push(if value { !0 } else { 0 });
+        }
+        v.len = len;
+        v.mask_last_block();
+        v
+    }
+}
+
+impl<T: Debug + PrimInt + One + Zero> Vob<T> {
+    /// Constructs a new, empty Vob (with a user-defined backing storage type) with the given
+    /// capacity.
+    ///
+    /// # Examples
+    /// ```
+    /// use vob::Vob;
+    /// let mut v = Vob::<u32>::new_with_storage_type(0);
+    /// v.push(true);
+    /// assert_eq!(v[0], true);
+    /// ```
+    pub fn new_with_storage_type(capacity: usize) -> Vob<T> {
+        Vob {
+            len: 0,
+            vec: Vec::with_capacity(capacity)
+        }
+    }
+
+    /// Returns the number of bits the Vob can hold without reallocating.
+    ///
+    /// # Examples
+    /// ```
+    /// use vob::Vob;
+    /// assert_eq!(Vob::new().capacity(), 0);
+    /// assert!(Vob::with_capacity(1).capacity() >= 1);
+    /// ```
+    pub fn capacity(&self) -> usize {
+        // This multiplication can't overflow because of the checks in reserve()
+        self.vec.capacity() * bits_per_block::<T>()
+    }
+
+    /// Reserves capacity for at least `additional` more bits to be inserted in the Vob. The Vob
+    /// may reserve more space to avoid frequent reallocations. After calling `reserve`, capacity
+    /// will be greater than or equal to `self.len() + additional`. Does nothing if capacity is
+    /// already sufficient.
+    ///
+    /// # Examples
+    /// ```
+    /// use vob::Vob;
+    /// let mut v = Vob::new();
+    /// v.reserve(1);
+    /// assert!(v.capacity() >= 1);
+    /// ```
+    pub fn reserve(&mut self, additional: usize) {
+        // Repeatedly calling this will over-reserve, since we only record how many bits we've
+        // used, not how many bits of capacity the user thinks we have. For example, repeatedly
+        // asking for storage for one extra bit will repeatedly add a whole extra machine word.
+        // Fixing this would require an extra machine word in the Vob struct, which hardly seems
+        // worth it for an unlikely, and probably silly, use case.
+        let cap = self.vec.capacity()
+                          .checked_mul(bits_per_block::<T>())
+                          .and_then(|x| x.checked_add(additional))
+                          .expect("Overflow detected");
+        self.vec.reserve(blocks_required::<T>(cap));
+    }
+
+    /// Shrinks the capacity of the vector as much as possible.
+    ///
+    /// It will drop down as close as possible to the length but the allocator may still inform the
+    /// vector that there is space for a few more elements.
+    pub fn shrink_to_fit(&mut self) {
+        self.vec.shrink_to_fit();
+    }
+
+    /// Shortens the Vob, keeping the first `len` elements and dropping the rest.
+    ///
+    /// If len is greater than the vector's current length, this has no effect.
+    ///
+    /// The drain method can emulate truncate, but causes the excess elements to be returned
+    /// instead of dropped.
+    ///
+    /// Note that this method has no effect on the allocated capacity of the vector.
+    ///
+    /// # Examples
+    /// ```
+    /// #[macro_use] extern crate vob;
+    /// fn main() {
+    ///     let mut v = vob![true, false, true];
+    ///     v.truncate(2);
+    ///     assert_eq!(v, vob![true, false]);
+    /// }
+    /// ```
+    pub fn truncate(&mut self, len: usize) {
+        if len > self.len {
+            return;
+        }
+        self.len = len;
+        self.vec.truncate(blocks_required::<T>(len));
+        self.mask_last_block();
+    }
+
+    /// Appends a bool to the back of the Vob.
+    ///
+    /// # Examples
+    /// ```
+    /// use vob::Vob;
+    /// let mut v = Vob::new();
+    /// v.push(true);
+    /// v.push(false);
+    /// assert_eq!(v.get(0), Some(true));
+    /// assert_eq!(v.get(1), Some(false));
+    /// ```
+    pub fn push(&mut self, value: bool) {
+        if self.len % bits_per_block::<T>() == 0 {
+            self.vec.push(T::zero());
+        }
+        let i = self.len;
+        self.len = i.checked_add(1).expect("Overflow detected");
+        self.set(i, value);
+    }
+
+    /// Removes the last element from the Vob and returns it, or `None` if it is empty.
+    ///
+    /// # Examples
+    /// ```
+    /// use vob::Vob;
+    /// let mut v = Vob::new();
+    /// v.push(true);
+    /// assert_eq!(v.pop(), Some(true));
+    /// assert_eq!(v.pop(), None);
+    /// ```
+    pub fn pop(&mut self) -> Option<bool> {
+        if self.len == 0 {
+            return None;
+        }
+        // The subtraction can't underflow because self.len > 0.
+        let v = self.get(self.len - 1);
+        debug_assert!(v.is_some());
+        self.len -= 1;
+        self.mask_last_block();
+        v
+    }
+
+    /// Returns the number of elements in the Vob.
+    pub fn len(&self) -> usize {
+        self.len
+    }
+
+    /// Returns true if the Vob has a length of 0.
+    ///
+    /// # Examples
+    /// ```
+    /// use vob::Vob;
+    /// assert_eq!(Vob::from_elem(2, true).is_empty(), false);
+    /// assert_eq!(Vob::new().is_empty(), true);
+    /// ```
+    pub fn is_empty(&self) -> bool {
+        self.len == 0
+    }
+
+    /// Splits the collection into two at the given index.
+    ///
+    /// Returns a newly allocated Self. self contains elements [0, at), and the returned Self
+    /// contains elements [at, len).
+    ///
+    /// Note that the capacity of self does not change.
+    ///
+    /// # Examples
+    /// ```
+    /// use vob::Vob;
+    /// let mut v1 = Vob::new();
+    /// v1.push(true);
+    /// v1.push(false);
+    /// let v2 = v1.split_off(1);
+    /// assert_eq!(v1, Vob::from_elem(1, true));
+    /// assert_eq!(v2, Vob::from_elem(1, false));
+    /// ```
+    pub fn split_off(&mut self, at: usize) -> Vob<T> {
+        if at >= self.len {
+            return Vob::<T>::new_with_storage_type(0);
+        }
+        let mut nv = Vob::<T>::new_with_storage_type(self.len - at);
+        // This could easily be made more efficient.
+        for blk in self.iter().skip(at) {
+            nv.push(blk);
+        }
+        self.len = at;
+        self.mask_last_block();
+        nv
+    }
+
+    /// Returns the value of the element at position `index` or `None` if out of bounds.
+    ///
+    /// # Examples
+    /// ```
+    /// use vob::Vob;
+    /// let mut v = Vob::new();
+    /// v.push(false);
+    /// assert_eq!(v.get(0), Some(false));
+    /// assert_eq!(v.get(1), None);
+    /// ```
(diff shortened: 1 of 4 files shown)


ltratt:
Here's an example of porting code from `BitVec` to `Vob` https://github.com/softdevteam/lrtable/commit/07f855a1db5bbee5d07dcdb767e1924059318482 (it wasn't difficult).

src/lib.rs
```diff
@@ -0,0 +1,1166 @@
...
+
+    /// Constructs a new, empty Vob (with `usize` backing storage, which is likely to be the best
+    /// choice in nearly all situations) with the specified capacity.
+    ///
+    /// The Vob will be able to hold at least `capacity` elements without reallocating. If
+    /// `capacity` is 0, the vector will not allocate.
+    pub fn with_capacity(capacity: usize) -> Vob<usize> {
+        Vob {
+            bits: 0,
+            vec: Vec::with_capacity(storage_size_align::<usize>(capacity))
```
vext01: So you round up to the next alignment boundary? Why?

src/lib.rs
```diff
@@ -0,0 +1,1166 @@
...
+            if ov != nv {
+                chngd = true;
+            }
+            *sb = nv;
+        }
+        self.mask_last_block();
+        chngd
+    }
+
+    /// We guarantee that the last storage block has no bits set past the "last" bit: this function
```
vext01: Why is this useful? This guarantee an optimisation?

src/lib.rs
```diff
@@ -0,0 +1,1166 @@
...
+/// Several functions in `BitVec` have different names in `Vob`, but porting is in general fairly
+/// simple. The main semantic difference is that `Vob`s [`clear()`](struct.Vob.html#method.clear)
+/// function empties the `Vob` of contents (i.e. sets its length to 0), whereas `BitVec`s function
+/// of the same name unsets all bits (keeping the length unchanged). `Vob`s equivalent is
+/// [`set_all(false)`](struct.Vob.html#method.set_all).
+#[derive(Clone)]
+pub struct Vob<T=usize> {
+    /// How many bits are stored in this Vob?
+    bits: usize,
+    /// The underlying storage. Since the storage consists of (potentially multiple-byte) blocks,
```
vext01: Can you define block somewhere? Is it one element in the underlying vec? e.g. `usize`?

src/lib.rs
```diff
@@ -0,0 +1,1166 @@
...
+    /// there may be "unused" bits in the final block. We guarantee that, at all points visible to
+    /// the user, the "unused" bits are set to 0.
+    vec: Vec<T>
+}
+
+// In an ideal world, Rust's type defaults would allow us to fold the two impl blocks into one and
+// say "impl Vob<T: usize>", but currently that doesn't work.
+impl Vob<usize> {
```
vext01: So we are specialising a handful of functions for `usize` backed vobs here? The other functions missing fall back on the generic versions?

src/lib.rs
```diff
@@ -0,0 +1,1166 @@
...
+///
+/// ## Storage backing type
+///
+/// `Vob`s default to using `usize` as a storage backing type. This is generally a substantial win
+/// over using smaller storage types if you use functions such as
+/// [`or()`](struct.Vob.html#method.or). In such cases, `usize` on a 64-bit machine is almost
+/// exactly twice as fast as using `u32`. If you only ever set and get individual bits, a smaller
+/// data type might be marginally more effective: for such use cases `u32` is around 1% faster than
+/// `usize` on a 64-bit machine. You can choose your own storage type with the
+/// [`new_with_storage_type()`](struct.Vob.html#method.new_with_storage_type) constructor. In
```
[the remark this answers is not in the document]
ltratt: Unfortunately Rust's default types don't really work as one would expect :( This means we end up with two `impl` blocks. The first `impl` block handles the "make `usize` the default" which contains a `new` function. The second `impl` block can't overload `new`, so has to use this awkward name.

In an ideal world we'd have had one `impl` block specified thus:

```rust
impl<T: Debug + PrimInt + One + Zero> Vob<T=usize> {
```

Unfortunately that just doesn't work in `rustc` at the moment. AFAICS the default types are a half-implemented hack...

src/lib.rs
```diff
@@ -0,0 +1,1166 @@
...
+/// `Vob` is directly inspired by `BitVec`, but aims to provide an interface aligned to `Vec<bool>`
+/// Several functions in `BitVec` have different names in `Vob`, but porting is in general fairly
+/// simple. The main semantic difference is that `Vob`s [`clear()`](struct.Vob.html#method.clear)
+/// function empties the `Vob` of contents (i.e. sets its length to 0), whereas `BitVec`s function
+/// of the same name unsets all bits (keeping the length unchanged). `Vob`s equivalent is
+/// [`set_all(false)`](struct.Vob.html#method.set_all).
+#[derive(Clone)]
+pub struct Vob<T=usize> {
+    /// How many bits are stored in this Vob?
+    bits: usize,
```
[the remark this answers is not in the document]
ltratt: Could just be `len`?

src/lib.rs
ltratt: Yes, a block == one int.

src/lib.rs
ltratt: Correct.

src/lib.rs
```diff
@@ -0,0 +1,1166 @@
...
+    /// The underlying storage. Since the storage consists of (potentially multiple-byte) blocks,
+    /// there may be "unused" bits in the final block. We guarantee that, at all points visible to
+    /// the user, the "unused" bits are set to 0.
+    vec: Vec<T>
+}
+
+// In an ideal world, Rust's type defaults would allow us to fold the two impl blocks into one and
```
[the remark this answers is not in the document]
ltratt: Not externally, IMHO. Most of the time "things just work" for users in the API. The chances of anyone needing `new_with_storage_type` are so low as not to be worth clogging up the docs IMHO.

src/lib.rs
ltratt: Otherwise you won't have enough space to store the bits.

src/lib.rs
ltratt: We allow users to iterate over the blocks: we don't want to see them bits set beyond `self.len()`, otherwise they'll get confused.

src/lib.rs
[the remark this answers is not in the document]
vext01: Fine.

src/lib.rs
vext01: Right, I thought so. Perhaps just add a sentence saying that's what a `block` is.

ltratt: Already pushed a commit which does so :)

src/lib.rs
vext01: Not sure I follow.

Suppose the use does:
```
let vob = vob::with_capacity(11);
```

Then you would actually allocate some larger amount than 11, yet 11 would have been enough?

src/lib.rs
vext01: I see. Next silly question, why expose block to the user?

src/lib.rs
ltratt: But you can't store 11 bits on their own: if you have 8 bit blocks, for example, you need two blocks to store 11 bits.

src/lib.rs
ltratt: Operating on blocks is *massively* faster than on individual bits. `Vob` provides `and`, `or`, and `xor` as common cases, but there are others. `cfgrammar` implements its own `intersect` which iterates over blocks, for example. [Should I include `intersect` in `Vob`? I'm not sure. It's difficult to know where to stop...]

vext01: I guess I would expect an interface which exposes a slice which cannot extend beyond the length of the vob. Might be hard to implement?

src/lib.rs
```diff
@@ -0,0 +1,1167 @@
...
+
+// This is based on the `BitVec` approach to indices. It's clearly a horrible way of doing things,
+// but until `IndexGet` is implemented, we're stuck.
+static TRUE: bool = true;
+static FALSE: bool = false;
+
+impl<T: Debug + One + PrimInt + Zero> Index<usize> for Vob<T> {
+    type Output = bool;
+
+    fn index(&self, index: usize) -> &bool {
```
vext01: Hah. So a reference received by this method can't observe a mutation?

src/lib.rs
ltratt: It's also convenient to maintain this guarantee internally, so it wouldn't really make sense to only do it in an interator.

src/lib.rs
ltratt: ?

ltratt:
Also pushed a commit to rename `bits` -> `len`. What would you like to do next?

vext01:
I'm going to have to set aside some time to review this properly, but probably not today. Is that OK?

src/lib.rs
vext01: Yeah, I think I'm talking garbage too.

There's no way to mutate a vector without going via a mutable reference, is there? Even if you own the vec.

ltratt: Correct.

ltratt:
There's about 3 other PRs blocked on this one...

ltratt:
Actually, I lie. I think only 1 is blocked (if I'm mildly creative).

src/lib.rs
vext01: Right, so disregard my comment :)

ltratt:
I wasn't clear: this PR can wait a bit if necessary.

vext01:
I'll prioritise this, but I need to get Tom's Krun feature done first.

src/lib.rs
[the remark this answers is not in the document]
vext01: I wonder if it would be better to force the user to specify a backing type rather than this `new_with_storage_type()` dance.

At first I was all like "huh, I wonder what storage type `0` is?", then i realised that it's just a workaround to allow us specify a different backing type and `0` is the capacity.

If we force the user to specify a type, then if and when Rust eventually allows inference on `new()`, it's a no-op for consumers to use the new interface. i.e. `let v: Vob<u8> = Vob::new()` will work now and when Rust is fixed.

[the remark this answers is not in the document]
ltratt: No, we *definitely* don't want them to specify it -- people will often choose an inappropriate type (for a good approximation of the truth `usize` is, AFAICS, always the right choice; but in case I'm wrong, I provide a backdoor). At some point Rust will fix default generic types and we can deprecate this method.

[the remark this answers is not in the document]
vext01: Right OK. It's not a nice interface :(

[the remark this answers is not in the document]
ltratt: What I could do (indeed, an earlier version did have) is a disclaimer on `new_with_storage_type` which says "unless you have a good reason for doing so, use `new` or `with_capacity`). [I worried this was a bit too pedantic, but maybe I was wrong.]

src/lib.rs
```diff
@@ -0,0 +1,1167 @@
...
+fn bits_per_block<T>() -> usize {
+    size_of::<T>() * 8
+}
+
+/// Return the offset in the vector of the storage block storing the bit `off`.
+fn storage_block_off<T>(off: usize) -> usize {
+    off / bits_per_block::<T>()
+}
+
+/// Takes as input a number of bits requiring storage; returns an aligned number of blocks needed
```
vext01: OK, I understand a comment you made on Jake's repo now.

I'd call this `pad_to_blocksize` or something, as it's clearer what the intention is.I thought you were optimising by ensuring the storage sits on an alignment boundary.

And, yes, I know wikipedia lists this use case as one of the three kinds of alignment :P

src/lib.rs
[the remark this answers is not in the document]
ltratt: [I also sort-of disagree that it's not a good interface: the main method `new` (and `with_capacity`) "do the right thing (TM)", but there's a backdoor (`new_with_storage_type`) forced upon us by a lack of functionality in rustc. The latter bit isn't ideal, but I wouldn't say it's bad. Certainly I've written many worse APIs, even in the last week!]

src/lib.rs
ltratt: `pad` to me suggests "actively change the size"... :/

vext01: Hrm, yeah?

OK, how about `storage_bytesize`.

vext01: That's not right. One moment.

vext01: `num_blocks_required`, perfect!

ltratt: `storage_blocks_required`?

vext01: sure!

ltratt: OK, commit with that change in pushed.

src/lib.rs
```diff
@@ -0,0 +1,1167 @@
...
+        self.mask_last_block();
+        chngd
+    }
+
+    /// We guarantee that the last storage block has no bits set past the "last" bit: this function
+    /// clears any such bits.
+    fn mask_last_block(&mut self) {
+        let db = self.len % bits_per_block::<T>();
+        if db > 0 {
+            let m = (T::one() << db) - T::one();
```
vext01: `m`, `b`, `v`?

`m` might be a mask?

I know you like short variable names, but a this is a little hard to read.

vext01: What does `db` stand for actually?

"do bits"?

ltratt: m = mask, b = block, v = value. I think the first two are, from context, pretty clear. The last one is, I admit, meaningless.

I also just realised that the three lines spell out "MBV" which is an amusing short version of a famous band ;)

ltratt: "dead bits". Should be "unused bits" or "ub". Will clarify.

src/lib.rs
```diff
@@ -0,0 +1,1174 @@
...
+    }
+}
+
+/// How many bits are stored in each underlying storage block?
+fn bits_per_block<T>() -> usize {
+    size_of::<T>() * 8
+}
+
+/// Return the offset in the vector of the storage block storing the bit `off`.
+fn storage_block_off<T>(off: usize) -> usize {
```
vext01: I feel like I'm being awkward here, but the name of this one is confusing too. I thought "block off" was intended as a verb, as in "to block off the storage" (like one "blocks off a road" for example), but it's actually meant to mean "offset".

ltratt: OK, I could call them `block_offset` and `blocks_required`? [I think `off` is a very common abbreviation for "offset", but maybe I'm wrong.]

src/lib.rs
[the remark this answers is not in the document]
vext01: To be clear. I think you've done the best with the limitations of the tools.

I'd also have accepted forcing a decision from the user, but we've already discussed that.

src/lib.rs
vext01: It is yes, it's just an unfortunate interaction with "block off".

`block_offset` could work, yeah. I can't think of anything better.

src/lib.rs
vext01: The problem I had in my head would only manifest if you could obtain a mutable reference to a bit in side a `Vob`. Do you see what I mean?

We don't support that, so it's moot.

You'd have to use a separate struct if you wanted such an interface:
```
struct MutBit<'v> {
    vob: &mut 'v Vob,
    bit: usize,
}

impl MutBit {
  fn set(self, val: bool) {
        ....
  }
}
```

src/lib.rs
```diff
@@ -0,0 +1,1175 @@
...
+/// For larger numbers of bits, Vobs can lead to a significant memory decrease and performance
+/// increase.
+///
+/// # Examples
+/// The `vob!` macro makes creating small `Vob`s easy:
+///
+/// ```rust
+/// # #[macro_use] extern crate vob;
+/// # fn main() {
+/// let mut v = vob![true, false, true];
```
vext01: Could this macro be updated to also support:
```
vob![1, 0, 0 , 0, 1, 1];
```

for example?

(Just a more concise way to make an initialised vob)

src/lib.rs
ltratt: Indeed, it's more than I want to support now.

src/lib.rs
ltratt: Not sure. Also not sure I want to think about it now ;) This notation only makes sense for very tiny Vobs anyway IMHO.

src/lib.rs
ltratt: Commit pushed.

src/lib.rs
vext01: True, it was just a passing idea. Let's leave it be.

src/lib.rs
vext01: Yeah, i'm not a huge fan of these one-letter variables, but if you must.

I think this function is good, have some free tests:
```
    #[test]
    fn test_mask_last_block1() {
        let mut v = vob![true, true];
        assert_eq!(v.vec, vec![3]);

        v.vec[0] = 0xaaaaaaaa;
        v.len = 7;
        v.mask_last_block();
        assert_eq!(v.vec, vec![42]);

        v.len = 30;
        v.vec[0] = 0xffffaaaa;
        v.mask_last_block();
        assert_eq!(v.vec, vec![1073719978]);
    }

    #[test]
    fn test_mask_last_block2() {
        let mut v = Vob::from_elem(128, true);
        let full_block = 0xffffffffffffffff;
        assert_eq!(v.vec, vec![full_block, full_block]);

        let one_zero = 0xaaaaaaaaaaaaaaaa;
        v.len = 68;
        v.vec[0] = one_zero;
        v.vec[1] = v.vec[0];
        v.mask_last_block();
        assert_eq!(v.vec, vec![one_zero, 0b1010]);
    }
```

ltratt: I think these tests assume `u64` backing?

vext01: yes, you'd need to polish them up if you wanted to integrate them.

I just wanted to be sure that this function did what it claimed and used this code.

ltratt: The first works with 32 bit storage; the second doesn't. Fixing...

src/lib.rs
```diff
@@ -0,0 +1,1175 @@
...
+    /// vector that there is space for a few more elements.
+    pub fn shrink_to_fit(&mut self) {
+        self.vec.shrink_to_fit();
+    }
+
+    /// Shortens the Vob, keeping the first `len` elements and dropping the rest.
+    ///
+    /// If len is greater than the vector's current length, this has no effect.
+    ///
+    /// The drain method can emulate truncate, but causes the excess elements to be returned instead of dropped.
```
vext01: Long line? Vim seems to want to wrap it around 100.

ltratt: Oops, quite right!

src/lib.rs
vext01: Ah, it just so happens I didn't touch the top 32 bits in the first, yes. It should work on both.

src/lib.rs
```diff
@@ -0,0 +1,1175 @@
...
+    /// let v2 = v1.split_off(1);
+    /// assert_eq!(v1, Vob::from_elem(1, true));
+    /// assert_eq!(v2, Vob::from_elem(1, false));
+    /// ```
+    pub fn split_off(&mut self, at: usize) -> Vob<T> {
+        if at >= self.len {
+            return Vob::<T>::new_with_storage_type(0);
+        }
+        let mut nv = Vob::<T>::new_with_storage_type(self.len - at);
+        // This could easily be made more efficient.
```
vext01: Indeed. Obvious first step is to copy all "complete" storage blocks using a slice clone, and special case the other remaining block, if there is one. Can you raise an issue?

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    /// assert_eq!(v.get(0), Some(true));
+    /// assert_eq!(v.set(0, false), Some(true));
+    /// assert_eq!(v.set(0, false), Some(false));
+    /// ```
+    pub fn set(&mut self, index: usize, value: bool) -> Option<bool> {
+        if index >= self.len {
+            return None;
+        }
+        let m = T::one() << (index % bits_per_block::<T>());
+        let b = block_offset::<T>(index);
```
vext01: So the **o**ffset is called `b` and the **b**lock is called `o`?

Are you sure we want these short variable names? Without decent names, I can only store about 2 or 3 in my head. They are really hard to read.

src/lib.rs
ltratt: I'd rather wait until someone complains (does anyone use `split_off`? I never have...). It's only inefficient if someone has a very particular use case, at which point we can actually benchmark things.

src/lib.rs
ltratt: This is a bad example, I admit.

ltratt: [i.e. I used the wrong names.]

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    pub fn get(&self, index: usize) -> Option<bool> {
+        if index >= self.len {
+            return None;
+        }
+        let b = self.vec[block_offset::<T>(index)];
+        Some(b & (T::one() << (index % bits_per_block::<T>())) != T::zero())
+    }
+
+    /// Sets the value of the element at position `index` or `None` if out of bounds. Returns
+    /// `true` if this led to a change in the underlying storage or `None` if out of bounds.
```
vext01: It would return `Some(true)` if the storage changed, right?

Why is it useful to know that? I guess it's the Rust way?

src/lib.rs
vext01: Okie dokie.

[Have you been hanging around with physicists again?] :microscope:

src/lib.rs
ltratt: It's not the Rust way; but it's surprisingly often that you can optimise based on "did I actually change this underlying thing or not" IME (I use this feature in other code FWIW).

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    /// ```
+    pub fn iter_unset_bits(&self) -> IterUnsetBits<T> {
+        IterUnsetBits {
+            vob: self,
+            range: 0..self.len
+        }
+    }
+
+    /// Return an iterator over the underlying storage blocks. The last block is guaranteed to have
+    /// "unused" bits (i.e. those past `self.len()`) set to 0.
```
vext01: I see, so perhaps this is why you have the invariant that the internal storage has `0`s at the top.

I wonder if you would get any meaningful speed up by relaxing this constraint (and having, internally, the top bits undefined), and only masking off when the user need to perceive the top bits as `0`s.

Thoughts?

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    /// assert_eq!(v.len(), 129);
+    /// assert_eq!(v.get(0), Some(false));
+    /// assert_eq!(v.get(128), Some(true));
+    /// ```
+    pub fn resize(&mut self, new_len: usize, value: bool) {
+        if new_len <= self.len {
+            self.truncate(new_len);
+            return
+        }
+        if value && self.len > 0 {
```
vext01: Here's another place with that invariant baked in. Hrm..

src/lib.rs
ltratt: I doubt we could measure the difference in any reasonable use case (but I don't know for sure).

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    ///
+    /// # Examples
+    /// ```
+    /// #[macro_use] extern crate vob;
+    /// fn main() {
+    ///     let mut v = vob![true];
+    ///     v.extend_from_slice(&vec![false, true]);
+    ///     assert_eq!(v, vob![true, false, true]);
+    /// }
+    pub fn extend_from_slice(&mut self, other: &[bool]) {
```
vext01: Could also be optimised, although a little trickier.

Easy case, the length of `self` is right up to capacity. Then it's just a matter of using the underlying `vec`s extend method.

[not for this PR]

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    /// ```
+    /// #[macro_use] extern crate vob;
+    /// fn main() {
+    ///     let mut v = vob![true, false, true];
+    ///     v.set_all(false);
+    ///     assert_eq!(v, vob![false, false, false]);
+    /// }
+    /// ```
+    pub fn set_all(&mut self, value: bool) {
+        for b in self.vec.iter_mut() {
```
vext01: Another optimisation candidate.

vext01: No, I'm wrong.

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+        self.mask_last_block();
+    }
+
+    /// For each bit in this Vob, `and` it with the corresponding bit in `other`, returning `true`
+    /// if this led to any changes or `false` otherwise. The two Vobs must have the same number of
+    /// bits.
+    ///
+    /// # Panics
+    ///
+    /// If the two Vobs are of different length.
```
vext01: Is that what we want?

Isn't it common to e.g. `or` a 64-bit value, by say `0x2` (which can be expressed in 1-byte).

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+        let mut chngd = false;
+        for (sb, ob) in self.vec.iter_mut().zip(other.vec.iter()) {
+            let ov = *sb;
+            let nv = ov & *ob;
+            if ov != nv {
+                chngd = true;
+            }
+            *sb = nv;
+        }
+        self.mask_last_block();
```
vext01: If the Vobs are the same length, they will both have the same number of `0`s at the top, and `0 & 0 == 0` so I think you can kill the masking? And in the `or` function too?

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+        write!(fmt, "Vob[")?;
+        for b in self {
+            write!(fmt, "{}", if b { 1 } else { 0 })?;
+        }
+        write!(fmt, "]")?;
+        Ok(())
+    }
+}
+
+impl<T: Debug + One + PrimInt + Zero> Extend<bool> for Vob<T> {
```
vext01: Neat!

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+
+    fn next(&mut self) -> Option<usize> {
+        if let Some(mut i) = self.range.next() {
+            for b in block_offset::<T>(i) .. blocks_required::<T>(self.vob.len) {
+                let v = self.vob.vec[b];
+                if v != T::zero() {
+                    let i_off = i % bits_per_block::<T>();
+                    let tz = (v >> i_off).trailing_zeros() as usize;
+                    if tz < bits_per_block::<T>() {
+                        let bs = b * bits_per_block::<T>() + i_off + tz;
```
vext01: `bs`?

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    range: Range<usize>,
+}
+
+impl<'a, T: Debug + One + PrimInt + Zero> Iterator for IterSetBits<'a, T> {
+    type Item = usize;
+
+    fn next(&mut self) -> Option<usize> {
+        if let Some(mut i) = self.range.next() {
+            for b in block_offset::<T>(i) .. blocks_required::<T>(self.vob.len) {
+                let v = self.vob.vec[b];
```
vext01: I'd be tempted to add some comments to this fiddly piece of logic here, but I won't insist.

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+                if v != T::zero() {
+                    let i_off = i % bits_per_block::<T>();
+                    let tz = (v >> i_off).trailing_zeros() as usize;
```
vext01: Just to check, this function is fully generic? It will work for other block types?

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+#[derive(Clone)]
+pub struct IterUnsetBits<'a, T: 'a> {
+    vob: &'a Vob<T>,
+    range: Range<usize>,
+}
+
+impl<'a, T: Debug + One + PrimInt + Zero> Iterator for IterUnsetBits<'a, T> {
+    type Item = usize;
+
+    fn next(&mut self) -> Option<usize> {
```
vext01: Would it be easier to implement this in terms of `IterSetBits`?

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+        if self.len != other.len {
+            return false;
+        }
+        self.iter_storage()
+            .zip(other.iter_storage())
+            .all(|(w1, w2)| w1 == w2)
+    }
+}
+
+impl<T: Debug + One + PrimInt + Zero> Eq for Vob<T> {}
```
vext01: I have to admit, I don't fully understand `Eq` and `PartialEq` in rust.

Looks to me like we have a total ordering here, so I'd expect to implement `Eq` not `PartialEq`, but `Eq` has no methods.

The docs didn't manage to explain it to me so well. I'll look into it more some time.

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+
+impl<T :Debug + Hash + One + PrimInt + Zero> Hash for Vob<T> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.len.hash(state);
```
vext01: Was there an informed decision behind including the length in the hash?

I can't decide if including it would lead to fewer collisions...

vext01:
That concludes my in-depth review. I learned a lot along the way!

src/lib.rs
ltratt: Agreed.

src/lib.rs
ltratt: This is what `BitVec` does and I think it's the right thing. It really doesn't make sense to `or` things of unequal length (what happens with different storage sizes?!).

src/lib.rs
ltratt: Good point.

src/lib.rs
ltratt: Point taken.

src/lib.rs
ltratt: Erp, good point. In practise, yes... but I agree, it doesn't look good. It might be difficult to do better though. Let me try.

src/lib.rs
ltratt: I don't think so (at least, I can't work out how!).

src/lib.rs
ltratt: OK, this is in fact safe; but we could probably use `u32` for `bits_per_block` as the `trailing_zeros` function does. That upsets a lot of code though because `num_bits` is `usize`, which means that any operation involving the two types requires `as`. I'm not sure it's worth fixing.

src/lib.rs
ltratt: I think I was influenced by `Vec`. But I agree: it does seem odd. Let's kill it, and if we've done something daft, someone will tell us...

src/lib.rs
ltratt: You have to implement `PartialEq` in order to be able to implement `Eq` (but the latter doesn't require additional methods). It is a bit weird, but it does eventually make sense.

ltratt:
OK, I *think* that addresses all your comments. Let me know if it doesn't!

src/lib.rs
vext01: Fine.

src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    /// let mut v = Vob::new();
+    /// v.push(false);
+    /// v.push(true);
+    /// let mut iterator = v.iter();
+    /// assert_eq!(iterator.next(), Some(false));
+    /// assert_eq!(iterator.next(), Some(true));
+    /// assert_eq!(iterator.next(), None);
+    /// ```
+    pub fn iter(&self) -> Iter<T> {
+        Iter {
```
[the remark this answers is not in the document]
vext01: ok

src/lib.rs
vext01: ok.

src/lib.rs
vext01: Well, if you had a `SetBitsIter` and it first gave you `6`, you'd know bits 1-5 are not set. then if it gave you `8`, you'd know `7` was not set.

It's just an idea. If it's not good, let's move on.

src/lib.rs
vext01: OK, you'll have to explain that one to me one day :)

src/lib.rs
vext01: ok

vext01:
All right. So unless you want to change the way the bits set iterator works, then I think this is ready for squashing.

Thanks for killing the one-letter variable names. I'm quite surprise I managed to sway you. How comes the change of heart? I think there's a sweet spot for variable names: we don't want `fractal_ratio_bias_after_normalisation`, but we don't want `f`.

I'll merge this first thing tomorrow so you are unblocked.

src/lib.rs
ltratt: Yes, but that's a declarative specification (actually, specification by example ;) ), not an algorithm. If you think about the work needed to turn it into an algorithm, it's clearly more work than the existing approach, because it's a superset of the work needed. Conclusion: I don't think it's a winner :)

src/lib.rs
ltratt: Basically they're special-cased in the compiler. Mostly it's not a very important detail IME.

ltratt:
Why did I change my mind? Weakness in my dotage...

Should I squash?

ltratt:
Oops, I just realised you meant I *should* squash if I thought now was the right time. So I've squashed -- ready to merge!

PR #3, opened 2018-02-01
crates.io still doesn't recognise the UPL so be explicit about it.

ltratt:
I have pushed a patch which, one day, should mean that crates.io will recognise the UPL. We're not there yet, unfortunately, so we need an explicit license file.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,31 @@
+Copyright (c) 2018 King's College London
+created by the Software Development Team <http://soft-dev.org/>
+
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -0,0 +1,31 @@
+Copyright (c) 2017 King's College London
```
vext01: Should that be 2018? When did you start working on this?

ltratt: Good spot! Fixed.

ltratt:
Squashed.

PR #4, opened 2018-03-13
Make iter_set_bits & iter_unset_bits take range arguments.

ltratt:
We will never be able to take a slice of a Vob, because a slice needs to be a (byte aligned) pointer. That means that the previous definitions of `iter_set_bits` and `iter_unset_bits` were only useful if you wanted to find either a) every set/unset bit b) the first N set/unset bits.

In particular, you could not say things like "find me all set/unset bits between positions 50 and 100". `iter_set_bits().skip(50).take(50)` might seem like it would work, but this would take the first 50 set bits, none of which might be between positions 50 and 100! Instead, one had to manually check each bit which is very slow.

This PR changes that: iter_set_bits and iter_unset_bits now take a range, like "drain". You can say things like `iter_set_bits(..)` or `iter_set_bits(10..)` and so on.

The bad news is that `RangeArgument` is unstable; that means we have to have our own version of the API in this crate. Since it's expected that Rust will eventually call this `RangeBounds`, I've chosen to use that terminology. This is a temporaryish hack until this feature is eventually stabilised in Rust, but it means that we can still run things on stable in the meantime.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,7 +37,7 @@
 
 extern crate num_traits;
 
-use std::cmp::PartialEq;
+use std::cmp::{min, PartialEq};
 use std::fmt;
 use std::fmt::Debug;
 use std::hash::{Hash, Hasher};
@@ -48,6 +48,11 @@ use std::slice;
 
 use num_traits::{PrimInt, One, Zero};
 
+// Whilst we wait for https://github.com/rust-lang/rust/issues/30877 to become stable, we can't use
+// RangeBounds and friends. We therefore have to implement a subset of the expected functionality
+// ourselves.
+mod range;
+use range::{Included, Excluded, RangeBounds, Unbounded};
 
 /// A Vob is a "vector of bits": a sequence of bits which exposes a `Vec`-like interface. Whereas
 /// `Vec<bool>` requires 1 byte of storage per bit, `Vob` requires only 1 bit of storage per bit.
@@ -76,7 +81,7 @@ use num_traits::{PrimInt, One, Zero};
 /// v2.set(67, true);
 /// v1.set(188, true);
 /// v1.and(&v2);
-/// let num_bits_set = v1.iter_set_bits().count();
+/// let num_bits_set = v1.iter_set_bits(..).count();
 /// assert_eq!(num_bits_set, 1);
 /// ```
 ///
@@ -424,8 +429,24 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
         }
     }
 
+    /// Convert a `RangeBounds` into a `Range`, taking into account this `Vob`'s length.
+    fn process_range<R>(&self, range: R) -> Range<usize>
+        where R: RangeBounds<usize>
+    {
+        let start = match range.start() {
+            Included(t) => min(*t, self.len),
+            Excluded(t) => min(*t + 1, self.len),
+            Unbounded => 0
+        };
+        let end = match range.end() {
+            Included(t) => min(*t + 1, self.len()),
+            Excluded(t) => min(*t, self.len()),
+            Unbounded => self.len
+        };
+        Range{start, end}
+    }
 
-    /// Returns an iterator which produces the index of each bit set in the Vob.
+    /// Returns an iterator which produces the index of each set bit in the specified range.
     ///
     /// # Examples
     /// ```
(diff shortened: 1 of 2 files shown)


src/range.rs
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2018 King's College London
```
vext01: Do we need the original Rust license in here too/instead?

src/range.rs
```diff
@@ -0,0 +1,79 @@
...
+// created by the Software Development Team <http://soft-dev.org/>
```
vext01: This whole file is verbatim modulo renamings?

src/lib.rs
```diff
@@ -823,9 +848,10 @@ impl<'a, T: Debug + One + PrimInt + Zero> Iterator for IterSetBits<'a, T> {
     type Item = usize;
 
     fn next(&mut self) -> Option<usize> {
+        debug_assert!(self.range.end <= self.vob.len);
```
vext01: Unrelated nitpick: In the line above, shouldn't we have `-> Option<Self::Item>` for maintainability sake?

src/lib.rs
```diff
@@ -837,7 +863,11 @@ impl<'a, T: Debug + One + PrimInt + Zero> Iterator for IterSetBits<'a, T> {
                         // There is another bit set after i in the block.
                         let bs = b * bits_per_block::<T>() + i_off + tz;
                         self.range.start = bs + 1;
-                        return Some(bs);
+                        if bs >= self.range.end {
```
vext01: Perhaps a separate issue, but the names `bs` and `tz` are not very descriptive. I'm thinking of timezones (`tz`) and, well `bs`, use your imagination :)

Closer inspection: `tz` is trailing zeroes,  but `bs` remains difficult to understand. What does it stand for?

I know I've reviewed this before, but as time passes I forget.

src/lib.rs
```diff
@@ -876,9 +907,9 @@ impl<'a, T: Debug + One + PrimInt + Zero> Iterator for IterUnsetBits<'a, T> {
                         // There is another bit unset after i in the block.
                         let bs = b * bits_per_block::<T>() + i_off + tz;
                         self.range.start = bs + 1;
-                        if bs >= self.vob.len {
-                            // This is the last block and the unset bit we thought we'd found is
-                            // actually an unused bit; we've thus reached the end of the Vob.
+                        if bs >= self.range.end {
```
vext01: Similarly here.

src/lib.rs
vext01: `b` is also not a good name. It could be `bits`, `bytes`, `blocks`...

In this instance I think it's a block.

src/range.rs
ltratt: Not quite, but not far off either. And I wasn't really sure what license to use. In retrospect, I might as well just use the Rust license in here.

src/lib.rs
ltratt: It will never be anything but `usize`: I'm not sure that the extra indirection helps users read the documentation?

src/lib.rs
ltratt: Whether they're good names or not, I don't think this PR is the right place to change them: it would lead to a lot of churn that isn't related to the PR. [FWIW, I found reading this bit of code fairly easy, and I can't say that of every bit of code I've written!]

ltratt: `bs` = "bit set position"

`tz` = "trailing zeros" (given that the line is `let tz = (v >> i_off).trailing_zeros() as usize;` I think this one is pretty clear ;) )

src/lib.rs
vext01: Up to you. I'd prefer to not duplicate the generic type myself.

ltratt: I'm neutral: you get a static type error if you change one but not the other. But I find rustdoc output harder to read when it says `next(&self) -> Option<Self::Item>` as there's now an extra level of indirection to understand...

src/lib.rs
vext01: As mentioned, I agree that this isn't the place for variable name changes above. But on this subject, (and probably a personal preference mismatch) I like variable names that don't force me to read too much surrounding code. If the variable name is obvious at a glance, then I can comprehend it quicker. I'd never have guessed  "bit set position" from `bs`, but `bit_pos` would have been excellent, and this conversation wouldn't be happening ;)

Anyway, resuming review.

src/lib.rs
vext01: I don't disagree :)

src/lib.rs
vext01: I'm trying to understand why we have ` + tz` in `let bs = b * bits_per_block::<T>() + i_off + tz;`.

I think I'm confused because I thought the bits would be stored "in order" even between blocks, e.g. if we have 8-bit blocks, the idices: `1 2 3 4 5 6 7 8 | 9 10 11 12...`, but I think they are really stored `8 7 6 5 4 3 2 1 | 16 15 14...`, which is why you have to skip over trailing zeroes, right?

ltratt: Your best bet is to run the code and put in print statements -- it is hard to reason about it statically if you're unfamiliar with the various tricks used here for performance. That said, the comment hopefully gives a good indication: `Find the next bit set after 'i % bits_per_block()'`. You'll also need to read up on `trailing_zeros`.

ltratt: [I can try an English explanation, though I don't know if it helps. "i_off is the last known set bit. tz_off is the number of unset bits to the next set bit. Add that to the overall block offset."]

vext01: OK, i misunderstood the meaning of `tz`. I thought it was the number of unallocated trailing zeros at the end of the block.

I'm thinking I'm doing a too in-depth review here. Your changes look fine.

vext01:
Ready to squash?

ltratt:
Squashed.

PR #5, opened 2018-03-13
Triple licence

ltratt:
This triple licenses Vob and prepares it for a 1.0.0 release.

[Warning: I'm not yet sure if crates.io accepts the "OR UPL-1.0" part, so I might end up making a second PR that moves us back to an explicit license file. I live in hope however!]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ repository = "https://github.com/softdevteam/vob/"
 version = "0.1.0"
 authors = ["Laurence Tratt <laurie@tratt.net>"]
 readme = "README.md"
-license-file = "LICENSE"
+license = "Apache-2.0 OR MIT OR UPL-1.0"
 categories = ["data-structures"]
 
 [dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "vob"
 description = "Vector of Bits with Vec-like API and usize backing storage"
 repository = "https://github.com/softdevteam/vob/"
-version = "0.1.0"
+version = "1.0.0"
 authors = ["Laurence Tratt <laurie@tratt.net>"]
 readme = "README.md"
 license = "Apache-2.0 OR MIT OR UPL-1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,4 +9,4 @@ license = "Apache-2.0 OR MIT OR UPL-1.0"
 categories = ["data-structures"]
 
 [dependencies]
-num-traits = "0.1.42"
+num-traits = "0.2.1"
--- a/LICENSE
+++ b/LICENSE
@@ -1,31 +1,4 @@
-Copyright (c) 2018 King's College London
-created by the Software Development Team <http://soft-dev.org/>
-
-The Universal Permissive License (UPL), Version 1.0
-
-Subject to the condition set forth below, permission is hereby granted to any person obtaining a
-copy of this software, associated documentation and/or data (collectively the "Software"), free
-of charge and under any and all copyright rights in the Software, and any and all patent rights
-owned or freely licensable by each licensor hereunder covering either (i) the unmodified
-Software as contributed to or provided by such licensor, or (ii) the Larger Works (as defined
-below), to deal in both
-
-(a) the Software, and
-(b) any piece of software and/or hardware listed in the lrgrwrks.txt file
-if one is included with the Software (each a "Larger Work" to which the Software is contributed
-by such licensors),
-
-without restriction, including without limitation the rights to copy, create derivative works
-of, display, perform, and distribute the Software and make, use, sell, offer for sale, import,
-export, have made, and have sold the Software and the Larger Work(s), and to sublicense the
-foregoing rights on either these or other terms.
-
-This license is subject to the following condition: The above copyright notice and either this
-complete permission notice or at a minimum a reference to the UPL must be included in all copies
-or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
-BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
-NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
-DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+Except as otherwise noted, this project is licensed under the Apache License,
+Version 2.0 <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0>, or
+the MIT license <LICENSE-MIT or http://opensource.org/licenses/MIT>, or the
+UPL-1.0 license <http://opensource.org/licenses/UPL> at your option.
--- a/README.md
+++ b/README.md
@@ -2,11 +2,36 @@
 
 A Vob is a "vector of bits": a sequence of bits which exposes a `Vec`-like
 interface. Whereas `Vec<bool>` requires 1 byte of storage per bit, `Vob`
-requires only 1 bit of storage per bit. For larger numbers of bits, Vobs can
-lead to a significant memory decrease and performance increase. `Vob` is broadly
+requires only 1 bit of storage per bit. `Vob` is broadly
 similar to [BitVec](https://crates.io/crates/bitvec-rs), but has an API more
 closely aligned to `Vec<bool>` and allows non-32-bit backing storage. On 64-bit
-systems this can lead to a substantial performance increase.
+systems this can lead to a substantial performance increase, particularly
+when functions such as
+[`iter_set_bits`](https://docs.rs/vob/0.1.0/vob/struct.Vob.html#method.iter_set_bits)
+are used.
 
-`Vob` currently doesn't implement all of the same functions one would find in
-`Vec<bool>`. We welcome contributions!
+
+## Migrating from `Vec<bool>`
+
+As far as possible, `Vob` is intended to have a superset of `Vec<bool>`'s interface, which
+should make porting most code fairly simple. However, `Vec<bool>` contains several functions
+which are not yet implemented in `Vob`: these are missing simply due to a lack of a current
+use-case rather than because of any fundamental incompatibilities.
+
+There is one missing feature which, currently, is impossible to implement: assignment to an
+index. In other words one cannot currently express `v[0] = true` for a `Vob` `v`. Until
+[`IndexGet` / `IndexMove` and equivalents](https://github.com/rust-lang/rfcs/issues/997) are
+implemented in `rustc`, this restriction appears to be inevitable. Note that referencing by
+index works (though via a hack identical to that used in `BitVec`): one can write
+`println!("{}", v[0])` for a `Vob` `v`, for example.
+
+
+## Migrating from `BitVec`
+
+`Vob` is directly inspired by the [`BitVec`](https://crates.io/crates/bit-vec) crate, but
+aims to provide an interface more closely aligned to `Vec<bool>` Several functions in
+`BitVec` have different names in `Vob`, but porting is in general fairly simple. The main
+semantic difference is that `Vob`s `clear()` function empties the `Vob` of contents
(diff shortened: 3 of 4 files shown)


ltratt:
Although it's not an obvious part of this PR, I've added a commit (#da7a8322) which gives slightly better motivation for a) why someone might want to switch to Vob b) why they should use `iter_set_bits` (and not e.g. `iter.filter()`).

LICENSE
```diff
@@ -1,31 +1,4 @@
...
-This license is subject to the following condition: The above copyright notice and either this
-complete permission notice or at a minimum a reference to the UPL must be included in all copies
-or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
-BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
-NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
-DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+Except as otherwise noted, this project is licensed under the Apache License,
```
vext01: Wouldn't it be an idea to show the licenses in full in this file? Web pages disappear.

Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "vob"
 description = "Vector of Bits with Vec-like API and usize backing storage"
 repository = "https://github.com/softdevteam/vob/"
-version = "0.1.0"
+version = "1.0.0"
```
vext01: Why not 0.2.0? Is this going to be the "we consider this done" release?

README.md
```diff
@@ -2,11 +2,36 @@
...
+when functions such as
+[`iter_set_bits`](https://docs.rs/vob/0.1.0/vob/struct.Vob.html#method.iter_set_bits)
+are used.
 
-`Vob` currently doesn't implement all of the same functions one would find in
-`Vec<bool>`. We welcome contributions!
+
+## Migrating from `Vec<bool>`
+
+As far as possible, `Vob` is intended to have a superset of `Vec<bool>`'s interface, which
```
vext01: Is "superset" correct here? I thought you had implemented only part of what `Vec` offers. In fact you say that in the next sentence.

README.md
```diff
@@ -2,11 +2,36 @@
...
+
+## Migrating from `BitVec`
+
+`Vob` is directly inspired by the [`BitVec`](https://crates.io/crates/bit-vec) crate, but
+aims to provide an interface more closely aligned to `Vec<bool>` Several functions in
+`BitVec` have different names in `Vob`, but porting is in general fairly simple. The main
+semantic difference is that `Vob`s `clear()` function empties the `Vob` of contents
+(i.e. sets its length to 0), whereas `BitVec`'s function of the same name unsets all bits
+(keeping the length unchanged). The same effect as `BitVec`'s `clear` can be achieved by
+using `Vob`'s `set_all(false)` function.
```
vext01: Nit: `set_all(false)` isn't a function.

src/lib.rs
```diff
@@ -93,13 +93,13 @@ use range::{Included, Excluded, RangeBounds, Unbounded};
...
-/// length to 0), whereas `BitVec`'s function of the same name unsets all bits (keeping the length
-/// unchanged). The same effect as `BitVec`'s `clear` can be achieved by using `Vob`'s
-/// [`set_all(false)`](struct.Vob.html#method.set_all) function.
+/// `Vob` is directly inspired by the [`BitVec`](https://crates.io/crates/bit-vec), but aims to
+/// provide an interface more closely aligned to `Vec<bool>` Several functions in `BitVec` have
+/// different names in `Vob`, but porting is in general fairly simple. The main semantic difference
+/// is that `Vob`s [`clear()`](struct.Vob.html#method.clear) function empties the `Vob` of contents
+/// (i.e. sets its length to 0), whereas `BitVec`'s function of the same name unsets all bits
+/// (keeping the length unchanged). The same effect as `BitVec`'s `clear` can be achieved by using
+/// `Vob`'s [`set_all(false)`](struct.Vob.html#method.set_all) function.
```
vext01: I wonder if the README could just point to the rustdoc, as we have duplication.

src/lib.rs
```diff
@@ -423,7 +423,9 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
         Range{start, end}
     }
 
-    /// Returns an iterator which produces the index of each set bit in the specified range.
+    /// Returns an iterator which efficiently produces the index of each set bit in the specified
+    /// range. Assuming appropriate support from your CPU, this is much more efficient than
```
vext01: Did you disassemble a binary to check that rustc uses vector instructions?

LICENSE
ltratt: I've stolen this from the way Rust does things. The Apache license is so long I wasn't sure if I wanted to use up so much space. But I can see your point. Let me have a quick think about it.

Cargo.toml
ltratt: It's a backwards incompatible change, so semantic versioning says that we have to do this.

README.md
ltratt: The weasel words are "intended to" ;)

src/lib.rs
ltratt: I did think about that, but I think it's good to see the "major" information when you land on the GitHub repo. Quite a lot of Rust crates do something similar.

LICENSE
vext01: It's only in one place, so I have no objection to including it in full.

Cargo.toml
vext01: Oh, I thought that was the second number, but you are right.

I've always considered PATCH and MINOR to be the same, but if that's how Rust does it, can't argue. Probably spent too much time with OpenBSD shared objects.

README.md
vext01: OK, as long as you think it's clear.

src/lib.rs
ltratt: I'm thinking of things like x86's `popcnt`: not every CPU, I think, has such an instruction. It makes things *much* faster if you have even mildly sparse Vobs. I don't know whether things get vectorised in this case: I doubt it very much, to be honest, as I suspect there's too much complex code going on for the compiler to manage to vectorise things.

Cargo.toml
ltratt: Rust crates are morally obliged to use semantic versinoing https://semver.org/ -- I'm starting to warm up to it, even though I prefer numbers that look more normal to humans.

README.md
ltratt: Let's pretend it's curried -- then it's a function ;)

Cargo.toml
vext01: Yeah, I knew the Rust guys dig semver, but I've never really looked into it. I'm not publishing crates you see ;)

src/lib.rs
vext01: OK.

src/lib.rs
vext01: So, I wonder why we are making these claims.

If you do a quick `objdump -D` and look at the code, you can dispel any doubt.

ltratt: Because x86 has `popcnt` and it makes some of my code 10x faster than doing `iter.filter()`.

ltratt: Think of it another way: if I didn't tell you that `iter_set_bits` was efficient in many cases, why would you bother using it?

LICENSE
ltratt: Aha, I think I now see why Rust doesn't include the text. Consider the standard MIT license text:

```
Copyright <YEAR> <COPYRIGHT HOLDER>
...
The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
```

If we expand this in the `LICENSE` file, it suggests everything in the crate has a single copyright holder, which is incorrect. But if we leave this on a per-file basis, everything makes sense. So simply deferring people to the website and individual files is quite a clever hack to avoid this problem.

In an ideal world, I think I'd probably drop the `LICENSE` file entirely, but I do appreciate that people often want a quick overview of the licence without wading through every source file.

Thoughts?

src/lib.rs
vext01: There seems to be confusion. I'm asking: are you certain rustc emitted `popcnt` into your binaries?

If so, great, this is OK. If not, then the docs are wrong.

ltratt: The doc comment makes no claims about `popcnt` -- I'm doing that to give you an intuition about why this stuff works :) Ultimately I depend entirely on what things like `trailing_zeros` in Rust's standard library use, which is why I added the "depending on processor support" qualification.

What I can tell you definitively is that I've measured `iter_set_bits` as being *much* faster than doing the semantically equivalent `iter().enumerate().filter(|(b, _)| b)`.

vext01: Saying it's faster is fine.

If we are certain that rustc uses the CPU extensions, then OK.

[FWIW, there's no use of popcnt in your rlib on release/openbsd]

ltratt: OK, I admit to taking a shortcut: `popcnt` is a shortcut for a family of instructions like `tzcnt` (which, I guess, is used somewhere!).

vext01: Afraid not:
```
arrakis:vob> for i in /usr/local/lib/rustlib/x86_64-unknown-openbsd/lib/*.so; do objdump -D $i | grep tzcnt; done
arrakis:vob> 
```

After some inspection, `trailing_zeroes` is implemented as a call to an LLVM intrinsic called `cttz`:
 * https://doc.rust-lang.org/src/core/num/mod.rs.html#1356-1358
 * https://github.com/rust-lang/rust/blob/master/src/librustc_trans/intrinsic.rs
 * https://llvm.org/docs/LangRef.html#llvm-cttz-intrinsic

So Rust has no explicit optimisations per-se. It's at the mercy of LLVM to do a good job. I couldn't find the implementation of `cttz`.

Unwinding, I'm less interested in talking about specific instructions, but I wonder if you should just say it's efficient and leave it at that?

vext01: Ah here's the implementation for `cttz` on x86_64 in LLVM 6 (IntrinsicLowering.cpp):
```
  case Intrinsic::cttz: {
    // cttz(x) -> ctpop(~X & (X-1))
    Value *Src = CI->getArgOperand(0);
    Value *NotSrc = Builder.CreateNot(Src);
    NotSrc->setName(Src->getName() + ".not");
    Value *SrcM1 = ConstantInt::get(Src->getType(), 1);
    SrcM1 = Builder.CreateSub(Src, SrcM1);
    Src = LowerCTPOP(Context, Builder.CreateAnd(NotSrc, SrcM1), CI);
    CI->replaceAllUsesWith(Src);
    break;
  }
```

But curiously there are no `ctpop` instructions in any of the libs shipped with Rust.

ltratt: Yes, it was always going to be at LLVM's mercy. But the point is that LLVM is then at your processor's mercy. If you don't have fast processor support for things like trailing_zeros, `iter_set_bits` won't be fast. So I can't make a guarantee about *all* platforms which is why I said "processor support".

vext01: I think you should remove the part about CPU features. Bencher8's CPU supports all of the features we are talking about, yet Rust isn't using them.

[If you disagree, then I'll fold, because this is taking too long at this point.]

ltratt: We've definitely gone too deep into this ;) but it turns out that on x86 LLVM is using the `bsf` instruction. I didn't know about that; it's a more generalised version of `popcnt` and the like. LLVM may also be using `bsf` because some recent x86 chips have a bug in `popcnt` (an incorrect dependency between the registers used) and friends that makes them slower.

Anyway, I think we've probably done this to death ;)

vext01: Now we are getting somewhere! All I've been asking is proof that your docs are accurate!

```
vext01@bencher8:~$ for i in ~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/*.so; do objdump -D $i | grep bsf; done
    47dc:	0f bc 76 a0          	bsf    -0x60(%rsi),%esi
    1220:	0f bc 67 a3          	bsf    -0x5d(%rdi),%esp
    4a19:	0f bc b1 af d4 e7 de 	bsf    -0x21182b51(%rcx),%esi
     d5e:	49 0f bc b7 c7 d7 d7 	bsf    0x3cd7d7c7(%r15),%rsi
    982d:	0f bc 8a f8 ec 86 d2 	bsf    -0x2d791308(%rdx),%ecx
  6bf73e:	0f bc ce             	bsf    %esi,%ecx
...
```

Given this, OK :)

vext01:
Ready to merge?

ltratt:
Yep!

PR #6, opened 2018-03-14
As feared, crates.io still doesn't know about UPL-1.0

ltratt:
Even though the actual license is Apache/MIT/UPL, we pretend to crates.io that it's just Apache/MIT. When, eventually, crates.io picks up the UPL, we can add that back in as another "OR". But I think simply dropping it from Cargo.toml is better than having the license show up as "non-standard".

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,9 @@ repository = "https://github.com/softdevteam/vob/"
 version = "1.0.0"
 authors = ["Laurence Tratt <laurie@tratt.net>"]
 readme = "README.md"
-license = "Apache-2.0 OR MIT OR UPL-1.0"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -5,7 +5,7 @@ repository = "https://github.com/softdevteam/vob/"
 version = "1.0.0"
 authors = ["Laurence Tratt <laurie@tratt.net>"]
 readme = "README.md"
-license = "Apache-2.0 OR MIT OR UPL-1.0"
```
vext01: Unlucky :) How about a comment to the effect of the PR description?

ltratt: Good idea.

ltratt:
Try that.

vext01:
LGTM. Squash away.

ltratt:
Squashed.

Issue #9, opened 2018-03-28
Serde support


Boscop:
Please add optional support for serde's Serialize/Deserialize traits.
Like this:
```toml
serde = { version = "1.0", features = ["derive"], optional = true }
```
```rust
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
struct Vob {
```

ltratt:
Good idea -- I'll look into this.

ltratt:
Now released as 1.1.0.

PR #10, opened 2018-03-28
Add Serde support.

ltratt:
This is enabled by an optional "serde" feature. It means that a program such as:

```rust
extern crate serde_json;
#[macro_use]
extern crate vob;

use vob::Vob;

fn main() {
    let mut v = vob![true, false, true];
    let serialized = serde_json::to_string(&v).unwrap();
    println!("{}", serialized);

    v.push(false);

    let deserialized: Vob = serde_json::from_str(&serialized).unwrap();
    println!("{:?}", deserialized);
}
```

prints out:

```
{"len":3,"vec":[5]}
Vob[101]
```

Addresses https://github.com/softdevteam/vob/issues/9

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "vob"
 description = "Vector of Bits with Vec-like API and usize backing storage"
 repository = "https://github.com/softdevteam/vob/"
-version = "1.0.0"
+version = "1.1.0"
 authors = ["Laurence Tratt <laurie@tratt.net>"]
 readme = "README.md"
 # This should be "Apache-2.0 OR MIT OR UPL-1.0" but crates.io doesn't know about
@@ -12,3 +12,4 @@ categories = ["data-structures"]
 
 [dependencies]
 num-traits = "0.2.1"
+serde = { version="1.0", features=["derive"], optional=true }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "vob"
 description = "Vector of Bits with Vec-like API and usize backing storage"
 repository = "https://github.com/softdevteam/vob/"
-version = "1.0.0"
+version = "1.0.1"
```
vext01: I'm not sure this is right:
```

    MAJOR version when you make incompatible API changes,
    MINOR version when you add functionality in a backwards-compatible manner, and
    PATCH version when you make backwards-compatible bug fixes.
```

You are adding features that don't break existing stuff, so you should bump minor, right?

vext01:
LGTM apart from that versioning query.

[Serde is pretty neat. I've been using it for a personal project recently]

Cargo.toml
ltratt: Good point!

ltratt:
OK fix pushed.

vext01:
Squash away.

ltratt:
Squashed.

Issue #11, opened 2018-04-09
Implement Vob::from_bytes


thomwiggers:
It'd be helpful to be able to create `Vob`s from `&[u8]` slices, as is supported in `BitVec`.

ltratt:
Agreed. It's a little (though not excessively) tricky to do because we really want to append a machine word at a time, which means a bit of fiddling around.

thomwiggers:
https://rust-bio.github.io/rust-bio/src/bit_vec/lib.rs.html#312-338 contains the BitVec implementation, which seems like it could probably be used as a basis.

I will probably send a PR some time once I get to the speed bottleneck FIXMEs in my project, if you don't get to it before me.

ltratt:
This feature is now in Vob 1.2.0.

PR #12, opened 2018-05-06
Add a from_bytes function.

ltratt:
The API (with the reversing of bits) is, intentionally, exactly the same as BitVec's. Our implementation is quite a bit different and, hopefully, a bit simpler.

Should address https://github.com/softdevteam/vob/issues/11 -- @thomwiggers is this the functionality you were asking for?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -138,7 +138,7 @@ impl Vob<usize> {
         }
     }
 
-    /// Creates a BitVec that holds `len` elements, setting each element to `value`.
+    /// Creates a Vob that holds `len` elements, setting each element to `value`.
     ///
     /// # Examples
     /// ```
@@ -154,6 +154,49 @@ impl Vob<usize> {
         }
         v.len = len;
         v.mask_last_block();
+        v
+    }
+
+    /// Create a Vob from a `u8` slice. The most significant bit of each byte comes first in the
+    /// resulting Vob.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// #[macro_use] extern crate vob;
+    /// use vob::Vob;
+    /// fn main() {
+    ///     let v = Vob::from_bytes(&[0b10100000, 0b00010010]);
+    ///     assert_eq!(v, vob![true, false, true, false, false, false, false, false,
+    ///                        false, false, false, true, false, false, true, false]);
+    /// }
+    /// ```
+    pub fn from_bytes(slice: &[u8]) -> Vob<usize> {
+        let mut v = Vob::with_capacity(slice.len()
+                                            .checked_mul(8)
+                                            .expect("Overflow detected"));
+        for i in 0..blocks_required::<usize>(slice.len() * 8) {
+            let mut w = usize::zero();
+            for j in 0..bytes_per_block::<usize>() {
+                let off = i * bytes_per_block::<usize>() + j;
+                if off >= slice.len() {
+                    // If slice doesn't contain a whole number of words, we'll end up with one
+                    // block at the end with "missing" bytes; since these are equivalent to 0, we
+                    // can simply ignore the fact they're missing.
+                    continue;
+                }
+                let b = slice[off];
+                if b != 0 {
+                    let mut rb: u8 = 0; // the byte b with its bits in reverse order
+                    for k in 0..8 {
+                        rb |= ((b >> k) & 1) << (8 - k - 1);
+                    }
+                    w |= (rb as usize) << (j * 8);
+                }
+            }
+            v.vec.push(w);
+        }
+        v.len = slice.len() * 8;
         v
     }
 }
@@ -766,7 +809,6 @@ impl FromIterator<bool> for Vob<usize> {
     }
 }
 
-
 // This is based on the `BitVec` approach to indices. It's clearly a horrible way of doing things,
 // but until `IndexGet` is implemented, we're stuck.
 static TRUE: bool = true;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -156,6 +156,51 @@ impl Vob<usize> {
...
+    ///     assert_eq!(v, vob![true, false, true, false, false, false, false, false,
+    ///                        false, false, false, true, false, false, true, false]);
+    /// }
+    /// ```
+    pub fn from_bytes(slice: &[u8]) -> Vob<usize> {
+        let mut v = Vob::with_capacity(slice.len()
+                                            .checked_mul(8)
+                                            .expect("Overflow detected"));
+        for i in 0..blocks_required::<usize>(slice.len() * 8) {
+            let mut w = usize::zero(); // word
```
vext01: Without sounding too pedantic, what value does the comment here add? I'd say it could even confusing, since a word is 16-bits on x86_64.

src/lib.rs
```diff
@@ -156,6 +156,51 @@ impl Vob<usize> {
...
+            for j in 0..bytes_per_block::<usize>() {
```
vext01: Maybe more semantic names for `i` and `j`?

vext01: `blk_idx`, `byte_idx`?

vext01: Similarly below, all of the one and two-letter variable names distract from comprehension. `i` `j` `b` `k` `rb` `w` `v` -- when did you become a physicist? ;)

src/lib.rs
```diff
@@ -156,6 +156,51 @@ impl Vob<usize> {
...
+                    // If slice doesn't contain a whole number of words, we'll end up with one
+                    // block at the end with "missing" bytes; since these are equivalent to 0, we
+                    // can simply ignore the fact they're missing.
+                    continue;
+                }
+                let b = slice[off];
+                if b != 0 {
+                    let mut rb: u8 = 0; // the byte b with its bits in reverse order
+                    for k in 0..8 {
+                        if b & (1 << k) != 0 {
```
vext01: This conditional is designed to avoid arithmetic and shifts if there is no need to flip a bit on?

Because otherwise, I think it's equivalent to:
```
for k in 0..8 {
    rb |= b & (1 << k) << (8 - k - 1);
}
```

src/lib.rs
```diff
@@ -952,7 +996,12 @@ impl<'a, T: Debug + One + PrimInt + Zero> Iterator for StorageIter<'a, T> {
 
 /// How many bits are stored in each underlying storage block?
 fn bits_per_block<T>() -> usize {
-    size_of::<T>() * 8
+    bytes_per_block::<T>() * 8
+}
+
+/// How many bytess are stored in each underlying storage block?
```
vext01: bytess

src/lib.rs
```diff
@@ -1095,6 +1144,29 @@ mod tests {
...
+                                  0b00110001, 0b10010101, 0b01111100, 0b01010001]);
+        assert_eq!(v, vob![true, false, true, false, false, false, false, false,
+                           false, false, false, true, false, false, true, false,
+                           false, false, true, true, false, true, false, true,
+                           true, true, false, false, true, false, true, false,
+                           false, false, true, true, false, false, false, true,
+                           true, false, false, true, false, true, false, true,
+                           false, true, true, true, true, true, false, false,
+                           false, true, false, true, false, false, false, true]);
+
```
vext01: blank lines, kill?

vext01: Test `Vob::from_bytes(&[])`? (empty byte vec).

ltratt:
I didn't know about `reverse_bits`, so that's interesting! I think I'd prefer to wait until that stabilises: feature gating this stuff gets a bit awkward :/

src/lib.rs
ltratt: Good spot.

src/lib.rs
ltratt: Agreed about testing the empty slice.

src/lib.rs
vext01: Entirely up to you. Maybe benchmark to see if one is faster?

vext01:
@thomwiggers Ah ha! I was looking for a function like that. There's a chance that it might translate to a fast machine instruction?

src/lib.rs
ltratt: It should be `rb |= ((b >> k) & 1) << (8 - k - 1);`. It's still going to be pretty fast, and it's shorter, so I like it.

src/lib.rs
ltratt: I must admit, I like the short names. But, for the record, I'm far too stupid to be a physicist.

src/lib.rs
ltratt: "Word" means "native integer size" pretty much everywhere except (maybe only in the past?) Windows AFAIK.

vext01: > "Word" means "native integer size"

Ah, if only it were that simple :) Brace yourself for pedantry!

I refer you to The "Intel® 64 and IA-32 Architectures Software Developer’s Manual
Volume 1:Basic Architecture":

x86_64

Just kill the comment. We all know that a `usize` is the size of a pointer, whereas "word" is not all that portable, sadly.

[Sorry to throw the book at you, especially since in this instance it's such a heavy book! In fact, the book comes in separate volumes, so I'd even have the opportunity to reload!]

vext01: So on x86_64, a `usize` is a quadword!

ltratt: Intel can define "the" to mean "spade", but I'll still call "the" "the". Wikipedia uses the same definition as I do ;) Anyway, I've killed the comment, but I now wonder how many conversations you and I have had where I've used "word" (==native integer size) and you've heard "16 bits"!

ltratt:
OK, see if that commit works for you.

src/lib.rs
vext01: I know what you mean, but more eyes than just ours will see this code. It's best to have precise comments, or let the code do the talking.

You've opted for the latter here, so I'm OK with that.

vext01:
OK, so do you want to squash.

ltratt:
Squashed.

thomwiggers:
> @thomwiggers Ah ha! I was looking for a function like that. There's a chance that it might translate to a fast machine instruction?

Yeah, it's supposed to use intrinsics where available (e.g. on ARM it should use `RBIT`).

PR #13, opened 2018-05-23
Allow accessing storage mutably

thomwiggers:
This allows me to directly write blocks as I'm reading them from my matrices' backing storage,
which are already in the right bit order.

I've marked it as unsafe, as it allows to invalidate the invariants of `Vob`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -804,9 +804,9 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
         chngd
     }
 
-    /// We guarantee that the last storage block has no bits set past the "last" bit: this function
-    /// clears any such bits.
-    fn mask_last_block(&mut self) {
+    // contents for the feature-guarded implementations next.
+    #[inline]
+    fn _mask_last_block(&mut self) {
         let ub = self.len % bits_per_block::<T>();
         // If there are no unused bits, there's no need to perform masking.
         if ub > 0 {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -759,6 +759,27 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
             }
         }
     }
+
+    /// Get a mutable reference to the underlying data structure
+    ///
+    /// This is marked as unsafe as it allows to invalidate the
```
vext01: Thanks for the PR. You should probably mention the assumptions that you ths user should not be violating in this comment? Wasn't there something about trailing zeros in storage blocks? IIRC Vob will assume that "unused" bits in the storage are zero.

thomwiggers: That's correct, and probably a good point. It may also be worthwhile exposing the cleanup function available to the public API, and also allowing to set the length.

vext01: Hrm, I'm not sure about that. Best wait for @ltratt to comment.

thomwiggers: FWIW, the `Vec` library also has `set_len()` and `as_mut_ptr`, that basically do the same.

ltratt:
This is an interesting idea but I think the fact that we have to document the current assumptions suggests that it's an even more dangerous API than it might first appear. I fear that it also fixes Vob's internal representation for ever because it exposes so much about the current implementation (including some details I'd probably prefer we didn't expose at all, if possible).

I guess my first question is: what are the use cases? Maybe we really do need to expose so much detail, but perhaps we can get away with being more abstract. For example, maybe we can provide a `mut_blocks` iterator that passes a block in and receives a block back; that way we can look after things like the Vob's length internally.

I must admit that I can't really see a good reason for `set_len`: either I want to shorten (`truncate`) or lengthen the Vob (admittedly `extend` doesn't quite do this easily, but we could easily add something like `push_n(bool, int)` which would change the Vob's length; if you did `push_n(false, some_number_that_doesn't_increase_the_blocks` it wouldn't even touch the blocks).

thomwiggers:
My use case is perhaps not the most common one, but one where I would greatly benefit from having access to these internals: I'm working with a library for binary matrices and as such have `BinVector` and `BinMatrix` types. For the matrix operations I'm using `m4ri`, a fast matrix library. I'm using Vob as the storage back-end for my `BinVector` type.

`m4ri` is a C library, so I'm often copying data to and from its data structures. For example, a matrix is a bunch of `int64_t`s per row, and then a bunch of rows. These rows can easily be translated to vectors as it'd just be copying the bytes. Having to do them bit-by-bit however is highly frustrating and very slow.

ltratt:
OK, so for copying data out I guess the existing `iter_storage` function is sufficient? If so, I wonder if we can have an `iter_storage_mut` function which would allow users to mutate data, but allow us to e.g. enforce that the last bits are set to 0? I admit that `iter_storage` does leak something about the internal representation, but it probably leaks about the minimum possible.

thomwiggers:
It don't think using an `iter_storage_mut` would allow to extend the number of elements...

ltratt:
We have the various `extend*` functions for adding data. Is there something extra that they could do? Admittedly, the `extend` functions present a "fully aligned" API to the outside world: do you ever do clever tricks adding "bits at a time" that would benefit from non-aligned access?

thomwiggers:
Yeah, in my case, converting a (1xn) M4RI Matrix to a Vob would mean flipping bits twice as m4ri also stores bit 0 in the LSB.

ltratt:
OK, I see where you're coming from. I'm still really nervous about this, because it exposes complexity to the external world and even just seeing those functions in docs is going to confuse / tempt some people. Maybe we could feature gate this which, presumably, would mean these functions wouldn't show up in the normal docs?

thomwiggers:
A feature flag would perhaps be an interesting solution. Maybe `unstable` or `danger_zone` or something? :stuck_out_tongue:

thomwiggers:
I've implemented `unsafe_internals`

ltratt:
OK, this looks like it's on the right tracks. Can we do something to reduce the duplication between the pub/notpub `mask_last_block` functions? Maybe there's some macro hackery or something similar?

PS: It's best to just keep adding new commits and not rebasing, because it makes it harder for me to review. I'll ask you to squash/rebase when we've finished (i.e. just before merging).

thomwiggers:
I don't know if macros will allow the amount of token tree hackery to switch on `pub`. I've moved the implementation part of the function to a `_mask_last_block`

ltratt:
I think that's a reasonable solution.

My last main suggestion is that we put all the `unsafe_internals` into their own `impl` so we have something like:

```rust
#[cfg(feature = "unsafe_internals")]
impl Vob {
    pub unsafe fn get_storage_mut(&mut self) -> &mut Vec<T> {
        &mut self.vec
    }
}
```

and so on (and presumably we need a similer `impl` for the private `mask` function?).

thomwiggers:
I've just kept the public and private `mask_internal_blocks` and the private `_mask` functions in the main `impl` block, as it otherwise becomes a bit cluttered imo. The other new public functions have been moved into their own `impl`

src/lib.rs
```diff
@@ -819,6 +819,91 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+    ///     assert_eq!(v1[0], true);
+    ///     assert_eq!(v1[1], false);
+    ///     assert_eq!(v1[2], false);
+    ///     assert_eq!(v1[8], true);
+    /// }
+    /// ```
+    pub unsafe fn set_len(&mut self, len: usize) {
+        self.len = len;
+    }
+
```
ltratt: Blank line.

ltratt:
A couple of stray blank lines, but nothing major. I also suspect we should edit `.travis.yml` to run the `unsafe_internal` tests on a nightly build?

thomwiggers:
Yeah I copied it out of `fast_reverse_bits`

ltratt:
OK. Can I ask: did you push a commit with the blank line fixes? Please don't force push if you can, as it makes my poor little brain hurt!

thomwiggers:
Maybe it's a suggestion to include `cargo fmt` into the (nightly) build pipeline...

thomwiggers:
The comment not being collapsed btw is a sign something still is an issue, normally. That one blank line slipped.

ltratt:
I'm happy for commits to just pile up: it doesn't worry me at all.

I'm not sure that running `cargo fmt` on Travis makes much difference, though?

ltratt:
I think we're ready for this to be squashed.

thomwiggers:
I've left two separate commits, the new build matrix justifies its own commit imo.

thomwiggers:
>I'm not sure that running `cargo fmt` on Travis makes much difference, though? 

It'll enforce its style guidelines, which will in turn get rid of those lines.

ltratt:
Thanks. Looks good.

PR #14, opened 2018-05-26
start/end are now start_bounds/end_bounds in Rust.

ltratt:
See https://github.com/rust-lang/rust/commit/07c415c2154f29d6dce8da0154ef41c8a5497fbf

I assume this is likely to be in Rust 1.27, so changing our implementation of it in preparation is a simple way of getting us ready. Assuming this does make it into 1.27, we can delete our version of range entirely.


vext01:
Can't really argue with that. Shall I merge?

ltratt:
Be my guest!

PR #15, opened 2018-05-28
Add extend_from_vob

thomwiggers:
Because this allows to use the assumptions in Vob, it's over a 100 times faster

Includes benchmarks

    running 4 tests
    test empty                  ... bench:          28 ns/iter (+/- 2)
    test extend                 ... bench:     469,014 ns/iter (+/- 60,205)
    test extend_vob_aligned     ... bench:         857 ns/iter (+/- 61)
    test extend_vob_not_aligned ... bench:       4,793 ns/iter (+/- 277)

### DIFF
--- a/benches/vob.rs
+++ b/benches/vob.rs
@@ -1,0 +1,51 @@
+#![feature(test)]
+
+#[macro_use]
+extern crate vob;
+extern crate test;
+
+use test::Bencher;
+use vob::*;
+
+const N: usize = 100000;
+
+#[bench]
+fn empty(b: &mut Bencher) {
+    b.iter(|| Vob::with_capacity(N));
+}
+
+#[bench]
+fn extend(b: &mut Bencher) {
+    // first initialise the source vector to drain
+    let mut source = Vob::with_capacity(N);
+    source.extend((0..N).map(|i| i % 2 == 0));
+
+    b.iter(|| {
+        let mut vector = Vob::with_capacity(N);
+        vector.extend(source.iter())
+    });
+}
+
+#[bench]
+fn extend_vob_aligned(b: &mut Bencher) {
+    // first initialise the source vector to drain
+    let mut source = Vob::with_capacity(N);
+    source.extend((0..N).map(|i| i % 2 == 0));
+
+    b.iter(|| {
+        let mut vector = Vob::with_capacity(N);
+        vector.extend_from_vob(&source)
+    });
+}
+
+#[bench]
+fn extend_vob_not_aligned(b: &mut Bencher) {
+    // first initialise the source vector to drain
+    let mut source = Vob::with_capacity(N);
+    source.extend((0..N).map(|i| i % 2 == 0));
+
+    b.iter(|| {
+        let mut vector = vob![true];
+        vector.extend_from_vob(&source)
+    });
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,8 @@ extern crate num_traits;
 #[cfg(feature = "serde")]
 #[macro_use]
 extern crate serde;
+#[cfg(test)]
+extern crate rand;
 
 use std::cmp::{min, PartialEq};
 use std::fmt;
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -1320,4 +1370,29 @@ mod tests {
         assert_eq!(v2, vob![2; true]);
         assert_ne!(v2, vob![2; false]);
     }
+
+    fn random_vector(len: usize) -> Vob {
```
ltratt: Shouldn't this be `random_vob`? Or maybe `rand_vob_data` or similar?

src/lib.rs
```diff
@@ -590,6 +592,52 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
         }
     }
 
+    /// Append all elements from a second Vob to this Vob.
```
ltratt: I think we need to say *why* you might want to use this function over just `extend` (saying something along the lines of "This is much faster than `extend`" is probably enough).

src/lib.rs
```diff
@@ -590,6 +592,52 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+    ///     let mut v1 = vob![true];
+    ///     let v2 = vob![false, false];
+    ///     v1.extend_from_vob(&v2);
+    ///     assert_eq!(v1, vob![true, false, false]);
+    /// }
+    /// ```    
+    pub fn extend_from_vob(&mut self, other: &Vob<T>) {
+        self.reserve(other.len());
+        let ub = self.len % bits_per_block::<T>();
+        // If there are no unused bits, we can just push the new blocks
```
ltratt: I think this comment should be inside the `if` (to match the comment in the `else` branch).

src/lib.rs
```diff
@@ -590,6 +592,52 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+        if ub == 0 {
+            self.vec.extend(other.vec.clone());
+        } else {
+            // we need to do things very carefully here.
```
ltratt: Can we format this comment more like others in the file? So wrapping to 100 chars, and using paragraphs etc.

src/lib.rs
```diff
@@ -590,6 +592,52 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+            // we need to shift each block ub to the left
+            // We use rotate to then add the bytes to the last block of this one
+            // We're relying on the bit ordering here.
+            // We're also relying on the last bits to be 0
+
+            // this mask must have the last ub bits not set
```
ltratt: Why not add a `debug_assert` instead of this comment? Much more effective IMHO!

benches/vob.rs
```diff
@@ -0,0 +1,54 @@
+#![feature(test)]
+#![feature(concat_idents_macro)]
```
ltratt: I don't think this feature is needed?

ltratt:
This looks like a good idea to me. I've made several minor comments. I'd also like to see the comments, in general, homogenised as far as possible with the rest of the file (which is annoying, I admit, but it makes long-term maintenance that bit easier). But I think we should be able to get this in fairly soon.

ltratt:
It also seems there's some trailing spaces in the patch that should be removed?

benches/vob.rs
thomwiggers: oh yeah, that got included from the benchmark I copied from another project.

src/lib.rs
thomwiggers: I've clarified this.

thomwiggers:
I have added a bunch of `debug_assert_eq!`s, formatted some stuff, added a few more comments and I think that should about wrap this up.

src/lib.rs
```diff
@@ -16,6 +16,8 @@ extern crate num_traits;
 #[cfg(feature = "serde")]
 #[macro_use]
 extern crate serde;
+#[cfg(test)]
+extern crate rand;
```
ltratt: Shouldn't this be in the `test` sub-module?

src/lib.rs
```diff
@@ -590,6 +592,64 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+            // left. We use rotate to move those bits to the bottom part of the integer. Then we
+            // add the bytes to the last block of this one.
+            //
+            // We're relying on the bit ordering here. We're also relying on the last bits to be 0.
+
+            // this mask has the last ub bits set
+            let msk = (T::one() << ub) - T::one();
+
+            for block in &other.vec {
+                //rotate block to the left
```
ltratt: (This is super nitpicky but...) Missing space after `//`.

src/lib.rs
```diff
@@ -590,6 +592,64 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+                let new_block: T = block.rotate_left(ub as u32);
+                {
+                    let last = self.vec.last_mut().unwrap();
+                    debug_assert_eq!(*last & !msk, T::zero(), "the last bits in the last block weren't 0");
+                    // add the last (upper) bits of new_block there
```
ltratt: Do we need "there" in this sentence? I think we can remove it.

ltratt:
OK, a few minor formatting comments, but nothing difficult.

src/lib.rs
thomwiggers: Crates need to be in lib right?

ltratt: Good point!

ltratt:
Looks good. Thanks!

PR #16, opened 2018-05-28
Use reverse_bits intrinsic when supported

thomwiggers:
This adds a new feature that allows to conditionally use the new
reverse_bits feature in Rust.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,11 @@
 language: rust
 
-matrix:
-  include:
-    - rust: nightly
-      env: FEATURES=unsafe_internals
-    - rust: nightly
-      env: FEATURES=''
-    - rust: stable
-    - rust: beta
+rust:
+  - stable
+  - nightly
+env:
+  - FEATURES=''
+  - FEATURES='unsafe_internals'
 
 before_script:
   - rustup component add rustfmt-preview
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ readme = "README.md"
 license = "Apache-2.0 OR MIT"
 categories = ["data-structures"]
 
+[build-dependencies]
+rustc_version = "0.2"
+
 [dependencies]
 num-traits = "0.2.1"
 serde = { version="1.0", features=["derive"], optional=true }
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,12 @@
+extern crate rustc_version;
+
+use rustc_version::{version_meta, Channel};
+
+fn main() {
+    // Set features depending on channel
+    if let Channel::Nightly = version_meta().unwrap().channel {
+        println!("cargo:rustc-cfg=nightly");
+        // Nightly supports https://github.com/rust-lang/rust/issues/48763
+        println!("cargo:rustc-cfg=reverse_bits");
+    }
+}
--- a/benches/vob.rs
+++ b/benches/vob.rs
@@ -101,3 +101,12 @@ fn and(bench: &mut Bencher) {
         a.and(&b);
     });
 }
+
+#[bench]
+fn from_bytes(b: &mut Bencher) {
+    let mut rng = rand::thread_rng();
+    let mut source = [0u8; 1024];
+    rng.fill(&mut source);
+
+    b.iter(|| Vob::from_bytes(&source));
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,4 @@
+#![cfg_attr(nightly, feature(reverse_bits))]
 // Copyright (c) 2018 King's College London created by the Software Development Team
 // <http://soft-dev.org/>
 //
@@ -162,6 +163,8 @@ impl Vob<usize> {
     /// Create a Vob from a `u8` slice. The most significant bit of each byte comes first in the
     /// resulting Vob.
     ///
+    /// If you are running nightly, this method will use the new `reverse_bits` intrinsic.
+    ///
     /// # Examples
     ///
     /// ```
@@ -187,12 +190,21 @@ impl Vob<usize> {
                     continue;
                 }
                 let b = slice[off];
-                if b != 0 {
-                    let mut rb: u8 = 0; // the byte b with its bits in reverse order
-                    for k in 0..8 {
-                        rb |= ((b >> k) & 1) << (8 - k - 1);
+                #[cfg(not(reverse_bits))]
+                {
+                    if b != 0 {
+                        {
+                            let mut rb: u8 = 0; // the byte b with its bits in reverse order
+                            for k in 0..8 {
+                                rb |= ((b >> k) & 1) << (8 - k - 1);
+                            }
+                            w |= (rb as usize) << (j * 8);
+                        }
                     }
-                    w |= (rb as usize) << (j * 8);
+                }
+                #[cfg(reverse_bits)]
+                {
+                    w |= (b.reverse_bits() as usize) << (j * 8);
                 }
             }
             v.vec.push(w);


ltratt:
Does this provide a noticeable speed-up?

ltratt:
Thinking about this a bit more: assuming this does give a useful speed-up (which I imagine it does, but we need to see numbers to be sure), wouldn't it be easier to use this automatically if `nightly` is used to build `vob`? Put another way: if it works well, I can't imagine anyone not wanting to use this feature. This would also make it completely transparent to the user, and mean that we don't have any future headaches removing a Cargo feature which some people have come to rely on.

thomwiggers:
> Does this provide a noticeable speed-up?
CI is running benchmarks when running builds on `nightly`; so you should be able to see the numbers there.

There is a slight speed-up on my system, but it should be more noticeable on e.g. ARMv8 where there is CPU support.

> wouldn't it be easier to use this automatically if nightly is used to build vob? Put another way: if it works well, I can't imagine anyone not wanting to use this feature. This would also make it completely transparent to the user, and mean that we don't have any future headaches removing a Cargo feature which some people have come to rely on.

Unfortunately, I don't know of a way to detect if a feature is available in 'normal' rust. There's ways to set a feature flag using a `build.rs`, but that seems a bit excessive. Also, once it stabilizes, you kind of still want to build with the feature flag turned on to opt in to the new behaviour - but detecting nightly would not be sufficient to do that anymore. 

See also https://internals.rust-lang.org/t/setting-cfg-nightly-on-nightly-by-default/1893

ltratt:
What sort of speed-up did you see? And, assuming it's an improvement, why would we want to make this an opt-in feature? It feels like it should be something the user doesn't know about: it just makes their code go faster.

thomwiggers:
4k ns to 3k or something.

It's not that I wouldn't want to, it's just that I can't figure out how to do it.

ltratt:
I must admit, for the complexity involved, I'm still wondering if we should just wait for this to stabilise. Admittedly, I'm not sure *when* it will stabilise.

ltratt:
In the interests of not letting things languish without a decision forever I think we should say that we won't merge this unless it can be detected automatically (with `build.sh` or some other trick). And if we don't want to do such a trick, it's probably better to close the PR. Sound like a plan?

thomwiggers:
That makes sense. Let's mark it WIP for now.

ltratt:
OK, I'll close this, but of course you can reopen it if/when automatic detection works.

thomwiggers:
Ok, I had some procrastination from writing on my master's thesis to do, so I implemented detection of nightly on top of the feature flags that are available.

There's now a `nightly` cfg flag that is automatically set on nightly. Its use is to enable the `#![feature]` tags. Those will only ever work on nightly, anyway.

The `nightly` flag is separate from the `reverse_bits` *cfg flag*. That flag will be set if nightly is detected and the `no_reverse_bits` feature is not enabled, or if the `reverse_bits` *feature* is enabled. `build.rs` contains a sanity check that both aren't enabled at the same time.

The `reverse_bits` feature will *not* switch on the `#[feature]` flag, to allow switching it on in `stable`. Of course, that won't work for the time being, but it will allow people that use Vob to opt-in to the new behaviour whenever https://github.com/rust-lang/rust/issues/48763 is stabilized, without a Vob update. (Though we should obviously still keep track of that issue)

By the way, I recommend switching on the `cron` settings of Travis to check nightly builds daily.

build.rs
```diff
@@ -0,0 +1,31 @@
...
+
+fn main() {
+    assert!(
+        !(env::var("CARGO_FEATURE_REVERSE_BITS").is_ok()
+            && env::var("CARGO_FEATURE_NO_REVERSE_BITS").is_ok()),
+        "Can't both enable and disable REVERSE_BITS"
+    );
+
+    // Set features depending on channel
+    match version_meta().unwrap().channel {
```
ltratt: Would this better be `if let Channel::Nightly = version_meta().unwrap().channel { ... } else { ... }`?

Cargo.toml
```diff
@@ -20,3 +23,4 @@ rand = "0.5"
 [features]
 unsafe_internals = []
 reverse_bits = []
```
ltratt: I don't think we need `reverse_bits` and `no_reverse_bits`. Personally I'd delete both, and just silently use reverse bits; but if we really, really want an option I think `no_reverse_bits` is enough.

ltratt:
Are we not able to query rustc directly for whether the `reverse_bits` features is available?

thomwiggers:
> Are we not able to query rustc directly for whether the reverse_bits features is available?

AFAIK, no. You could perhaps trail-compile something, but that seems way too much. That's why I included the `reverse_bits` feature.

ltratt:
@vext01 Do you know if you can query rustc for a feature directly?

@thomwiggers If not, I agree your approach is reasonable.

vext01:
You mean like:

```
#[cfg(feature = "foo")]
```
?

https://doc.rust-lang.org/1.26.0/book/first-edition/conditional-compilation.html

ltratt:
Not exactly. We'd like to detect when this feature https://doc.rust-lang.org/std/primitive.isize.html#method.reverse_bits is available or not (at the moment we assume "if we're on nightly, it's available", which might be a bit fragile).

vext01:
What do you want to do if it's not available?

ltratt:
If it's not available, we have a fallback (slower) route that doesn't involve using the feature.

vext01:
OK, so it would look like this I think:
```
#[cfg(feature = "reverse_bits")] {
    // fast code
}
#[cfg(not(feature = "reverse_bits"))] {
    // slow code
}
```

thomwiggers:
That only works for `Cargo.toml`-defined features.

vext01:
Ugh. That being the case, no idea then. Sorry.

vext01:
(I reckon that is a bug in rustc, why should compiler features be special?)

vext01:
The only other thing you can do is make a feature check in a build.rs, but that's horrible too.

ltratt:
OK, then I think Thom's suggestion is as good as it gets. We can live with it. There are still a couple of other comments for @thomwiggers though :)

.travis.yml
```diff
@@ -1,13 +1,27 @@
 language: rust
 
+rust:
+  - stable
+  - beta
```
ltratt: I must admit, I'm not very sure it's useful for us to test on `beta`. I know it doesn't *hurt*, but I feel a bit bad for using electricity for something that's probably never going to show an error.

thomwiggers: Seems like a good way to eliminate a lot of builds as well

thomwiggers: done.

Cargo.toml
thomwiggers: deleted the features.

build.rs
thomwiggers: done, especially since we no longer handle the feature settings.

Cargo.toml
```diff
@@ -19,3 +22,5 @@ rand = "0.5"
 
 [features]
 unsafe_internals = []
+reverse_bits = []
```
ltratt: Are these two lines supposed to be deleted?

thomwiggers: hmm somehow I missed them

ltratt: I thought you might be testing me :p

ltratt:
OK, I think this is ready to be squashed.

thomwiggers:
It's been squashed by the way

build.rs
```diff
@@ -0,0 +1,19 @@
+extern crate rustc_version;
+
+use rustc_version::{version_meta, Channel};
+use std::env;
+
+fn main() {
+    assert!(
+        !(env::var("CARGO_FEATURE_REVERSE_BITS").is_ok()
```
thomwiggers: oh wait, this needs to go.

thomwiggers: done.

ltratt:
Thanks!

PR #17, opened 2018-05-30
Don't reserve new blocks if capacity is sufficient

thomwiggers:
By checking the current length we'll avoid doing many mallocs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -247,16 +247,16 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
     /// assert!(v.capacity() >= 1);
     /// ```
     pub fn reserve(&mut self, additional: usize) {
-        // Repeatedly calling this will over-reserve, since we only record how many bits we've
-        // used, not how many bits of capacity the user thinks we have. For example, repeatedly
-        // asking for storage for one extra bit will repeatedly add a whole extra machine word.
-        // Fixing this would require an extra machine word in the Vob struct, which hardly seems
-        // worth it for an unlikely, and probably silly, use case.
-        let cap = self.vec.capacity()
-                          .checked_mul(bits_per_block::<T>())
-                          .and_then(|x| x.checked_add(additional))
-                          .expect("Overflow detected");
-        self.vec.reserve(blocks_required::<T>(cap));
+        let unused_bits: usize = self.capacity() - self.len();
+        if additional > unused_bits {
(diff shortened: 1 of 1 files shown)


thomwiggers:
Without this, this snippet of code will quickly use up all available RAM:

```
let v = Vob::with_capacity(10);
for _ in 0..100 {
    v.clear();
    v.extend_from_vob(&Vob::from_elem(3, false));
}

File: src/lib.rs [hunk 1]
```diff
@@ -247,16 +247,18 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
-        // used, not how many bits of capacity the user thinks we have. For example, repeatedly
-        // asking for storage for one extra bit will repeatedly add a whole extra machine word.
-        // Fixing this would require an extra machine word in the Vob struct, which hardly seems
-        // worth it for an unlikely, and probably silly, use case.
-        let cap = self.vec.capacity()
-                          .checked_mul(bits_per_block::<T>())
-                          .and_then(|x| x.checked_add(additional))
-                          .expect("Overflow detected");
-        self.vec.reserve(blocks_required::<T>(cap));
+        let current_free_space: usize = self.capacity() - self.len();
```
ltratt: Should we call this something shorter like `unused_bits`? Or even just `ub` as in one of the other functions.

src/lib.rs
```diff
@@ -247,16 +247,18 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+        if current_free_space > additional {
```
ltratt: We can simplify this to:

```rust
if additional > ub {
    // what is currently the else branch
}
```

ltratt:
A couple of minor comments, but we should be able to get this in quickly. I'm a bit disappointed I didn't think of doing this originally (instead of writing a long comment explaining why it wasn't worth doing...)!

src/lib.rs
thomwiggers: I've done `unused_bits`, as `ub` turned out to mean `used_bits` in most places I checked. It's sometimes a bit confusing as doing all this bit-mangling gives me headaches.

thomwiggers:
This one also already got fixed up, by the way.

ltratt:
If you squash this as you feel appropriate, I can merge it.

thomwiggers:
Rebased.

ltratt:
Thanks!

PR #18, opened 2018-05-31
Speed up split_off by copying entire Ts.

thomwiggers:
`split_off` can use a similar tactic as `extend_from_vob`. To that end I've refactored the
`extend_from_vob` into a private helper function that accepts an offset. This function is
then used by `split_off` to copy the majority of the upper bits into the new Vob.

The speedup is similar to the `extend_from_vob` speedup.

Also fixes a bug in `push` by resolving some assumption problems.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -306,6 +306,7 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
     /// assert_eq!(v.get(1), Some(false));
     /// ```
     pub fn push(&mut self, value: bool) {
+        debug_assert_eq!(self.vec.len(), blocks_required::<T>(self.len));
         if self.len % bits_per_block::<T>() == 0 {
             self.vec.push(T::zero());
         }
@@ -331,8 +332,8 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
         // The subtraction can't underflow because self.len > 0.
         let v = self.get(self.len - 1);
         debug_assert!(v.is_some());
-        self.len -= 1;
-        self.mask_last_block();
+        let new_len = self.len - 1;
+        self.truncate(new_len);
         v
     }
 
@@ -1503,5 +1504,31 @@ mod tests {
             storage[0] = 0b111;
         }
         assert_eq!(v1.get(1), Some(true));
+    }
+
+    #[test]
+    fn test_split_off() {
+        for len_a in 0..128 {
+            for len_b in 0..128 {
+                let a = random_vob(len_a as usize);
+                let b = random_vob(len_b as usize);
+                let mut joined = a.clone();
+                joined.extend_from_vob(&b);
+                assert_eq!(joined.len(), len_a + len_b);
+                let b_ = joined.split_off(len_a as usize);
+                assert_eq!(a, joined, "lower part for {}, {}", len_a, len_b);
+                assert_eq!(b, b_, "upper part for {}, {}", len_a, len_b);
+            }
+        }
+    }
+
+    #[test]
+    fn push_adjusts_vec_correctly() {
+        let mut v = Vob::new();
+        v.push(false);
+        assert_eq!(v.vec.len(), 1);
+        v.pop();
+        v.push(true);
+        assert_eq!(v.vec.len(), 1);
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,7 +24,7 @@ use std::fmt;
 use std::fmt::Debug;
 use std::hash::{Hash, Hasher};
 use std::iter::FromIterator;
-use std::mem::size_of;
+use std::mem::{replace, size_of};
 use std::ops::{Index, Range};
 use std::slice;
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -306,6 +306,7 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
     /// assert_eq!(v.get(1), Some(false));
     /// ```
     pub fn push(&mut self, value: bool) {
+        debug_assert_eq!(self.vec.len(), blocks_required::<T>(self.len));
```
thomwiggers: This assumption is used in several places, but I found out it's not really enforced correctly: `push` broke it. I've added these `debug_assert_eq`s in various places to make sure we catch other cases where it is violated.

src/lib.rs
```diff
@@ -331,8 +332,8 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
         // The subtraction can't underflow because self.len > 0.
         let v = self.get(self.len - 1);
         debug_assert!(v.is_some());
-        self.len -= 1;
-        self.mask_last_block();
+        let new_len = self.len - 1;
+        self.truncate(new_len);
```
thomwiggers: Doing `pop` repeatedly would leave entire blocks in `self.vec`.

src/lib.rs
```diff
@@ -306,6 +306,7 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
     /// assert_eq!(v.get(1), Some(false));
     /// ```
     pub fn push(&mut self, value: bool) {
+        debug_assert_eq!(self.vec.len(), blocks_required::<T>(self.len));
         if self.len % bits_per_block::<T>() == 0 {
             self.vec.push(T::zero());
```
thomwiggers: this line would break if you do:

```rust
let v = Vob::new();
v.push(false);
v.pop(); // now self.vec.len() == 1, but self.len() == 0
v.push(true); // now self.vec.len() == 2
v.pop() // ==> Some(false)
```

src/lib.rs
ltratt: Good idea.

src/lib.rs
ltratt: Oh, interesting! Can we break this out into its own test-case and commit? [Doesn't have to be a separate PR.]

src/lib.rs
```diff
@@ -608,12 +639,25 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
     /// ```
     pub fn extend_from_vob(&mut self, other: &Vob<T>) {
+        self.copy_into_with_offset(other, 0);
+    }
+
+    /// Copy blocks from other into self, respecting offset.
+    ///
+    /// block_offset * bits_per_block should be less than other.len
+    #[inline]
+    fn copy_into_with_offset(&mut self, other: &Vob<T>, block_offset: usize) {
```
ltratt: Maybe this could be something snappier like `extend_blocks`?

src/lib.rs
```diff
@@ -608,12 +639,25 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+        debug_assert!(
```
ltratt: I think we can just simplify this to `debug_assert!(block_offset * bits_per_block::<T>() <= other.len())` (not sure we need the message).

src/lib.rs
```diff
@@ -624,12 +668,16 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+            for block in other.vec.iter().skip(block_offset) {
                 // rotate block to the left
                 let new_block: T = block.rotate_left(ub as u32);
                 {
                     let last = self.vec.last_mut().unwrap();
-                    debug_assert_eq!(*last & !msk, T::zero(), "the last bits in the last block weren't 0");
+                    debug_assert_eq!(
+                        *last & !msk,
+                        T::zero(),
+                        "the last bits in the last block weren't 0"
```
ltratt: I think this message is fairly obvious so we can delete it.

src/lib.rs
thomwiggers: The doctest for `push` already triggered the `debug_assert_eq!`, but I've still added the testcase.

ltratt:
OK, I think we're close. Can we get the fix for https://github.com/softdevteam/vob/pull/18/commits/481321f48b43f6ca0685089d1eb47a19ec6ee051 into that commit too? That way we have a "here's a bug, a test case which shows the bug, and the fix for it" all in one (which makes it easier to track in the future). I appreciate that this will require some rebasing. If you want, you can also squash the "feedback" commit at the same time.

thomwiggers:
Into which commit?

ltratt:
You should feel free to squash things as you think appropriate: what we want to end up with is a nice, minimal, set of commits that looked like they were our intention all along :)

ltratt:
I think you need to rebase against master after the recent merge.

thomwiggers:
:heavy_check_mark:

ltratt:
Unfortunately you'll need to rebase again after I merged the other PR!

thomwiggers:
done, btw

PR #19, opened 2018-06-04
Run cargo fmt as part of CI

thomwiggers:
Also formats the code.

Only run `cargo fmt` on nightly, as there may be differences between versions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -174,8 +174,9 @@ impl Vob<usize> {
     /// }
     /// ```
     pub fn from_bytes(slice: &[u8]) -> Vob<usize> {
-        let mut v = Vob::with_capacity(slice.len().checked_mul(8).expect("Overflow detected"));
-        for i in 0..blocks_required::<usize>(slice.len() * 8) {
+        let new_len = slice.len().checked_mul(8).expect("Overflow detected");
+        let mut v = Vob::with_capacity(new_len);
+        for i in 0..blocks_required::<usize>(new_len) {
             let mut w = usize::zero();
             for j in 0..bytes_per_block::<usize>() {
                 let off = i * bytes_per_block::<usize>() + j;
@@ -196,7 +197,7 @@ impl Vob<usize> {
             }
             v.vec.push(w);
         }
-        v.len = slice.len() * 8;
+        v.len = new_len;
         v
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,13 +28,13 @@ use std::mem::{replace, size_of};
 use std::ops::{Index, Range};
 use std::slice;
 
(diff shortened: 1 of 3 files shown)


thomwiggers:
whoops, I branched in the wrong place.

src/lib.rs
```diff
@@ -552,7 +551,7 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
     pub fn resize(&mut self, new_len: usize, value: bool) {
         if new_len <= self.len {
             self.truncate(new_len);
-            return
+            return;
```
ltratt: This looks like it's a bug to me! Can we find a test case which triggers it?

ltratt:
My problem with `rustfmt` is that it makes several parts of the code harder to read :/ I am not entirely convinced by this, to be honest. Maybe we could take in *some* of the formatting suggestions?

src/lib.rs
thomwiggers: Pretty sure both `{ return }` and `return;` are type `()`

ltratt: Good point.

thomwiggers:
There was one part where there was a particularly big oneliner that I found particularly ugly, the other parts I was mostly fine with. That oneliner I moved an inner calculation into its own variable (which in turn allowed me to re-use its result a few time).

ltratt:
Why don't we cherry pick https://github.com/softdevteam/vob/pull/19/commits/2e6aec1806c0095615fe9c7f41427e37576b55b8 and https://github.com/softdevteam/vob/pull/19/commits/4344ed87750e5be7f1c5a02f683f7f234f87cd02, which are both pretty uncontroversial? We can then maybe think about the other things at some point in the future.

ltratt:
Once we've got those in, I think it's time for a 1.3.0 release. We can think about some of the other stuff for the next release.

ltratt:
OK, I think I've changed my mind on this. I still think rustfmt makes code uglier, but I also think it's clear that it's going to become the standard, and I'm not really very good at holding back tides. So, can we tidy this PR up into 3 commits along the lines of:

 * Splitting-one-line-into-a-variable
 * Running rustfmt over the whole codebase
 * Adding a rustfmt check to travis

and then I think we can probably merge this. Feel free to rebase / squash when doing this if it's easier (which I expect it will be).

thomwiggers:
I think I already had them like that if I look at the diff. I rebased on master though.

ltratt:
Can we retitle a couple of the commits? The first should probably have "fmt" -> "rustfmt"; the second maybe something like "Run rustfmt on existing code"?; and the third "split up ..." -> "Split overly dense line.".

thomwiggers:
Fair enough, done.

ltratt:
Thanks!

PR #22, opened 2018-06-06
Some formatting fixes in the source code

thomwiggers:
I've run rustfmt on the source code and selected the formatting fixes that
made sense to me. I've tried to be a bit conservative :smile:.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,13 +28,13 @@ use std::mem::{replace, size_of};
 use std::ops::{Index, Range};
 use std::slice;
 
-use num_traits::{PrimInt, One, Zero};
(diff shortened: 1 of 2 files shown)


ltratt:
There's still quite a bit in here that, to be honest, I'm not a fan of. e.g. what's the point of the churn adding `,` to the end of lines that don't need it? I'm happy to take the spilt-the-long-line-with-a-new-variable, and the ordering of imports alphabetically makes sense, but after that it just turns into a matter of personal taste. And, mostly, I'm not a fan of rustfmt's defaults (sorry).

thomwiggers:
Closed this in favour of #22

PR #24, opened 2018-06-20
Make new_with_storage_type create Vec with correct capacity

thomwiggers:
It is currently creating a greatly oversized `Vec`, as it is
directly taking the given capacity (supposedly bits) but reserving
for that many Ts.


ltratt:
Oops -- good spot!

PR #25, opened 2018-06-20
Don't mask in xor, it's not necessary

thomwiggers:
0 ^ 0 = 0

Also correct the comment in `or`, which is based on an incorrect assumption.


ltratt:
Agreed. I'm going to make a 1.3.1 release with the `new_with_storage_type` fix before I merge this in.

thomwiggers:
Done. I noticed I messed up my thomwiggers/master branch, so I also rebased #27 because this change got included in it as a result.

PR #26, opened 2018-06-20
Bump version to 1.3.1.

ltratt:
@thomwiggers I've sent a collaborator request to you for this repo, if you're willing. That way I can assign PRs like this one to you so that you can review and, if they're good enough quality, eventually merge (after reviewing and squashing etc).

This PR bumps Vob to 1.3.1 to include the `new_with_storage_type` fix which is harmless, but useful to anyone who's using Vob 1.3.0.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+# vob 1.3.1 (2018-06-20)
+
+* Don't overallocate memory in `new_with_storage_type`.
+
 # vob 1.3.0 (2018-06-11)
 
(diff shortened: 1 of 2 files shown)


thomwiggers:
Do you want to include #27 as well? 

(as a side note, I'm including my own branch in my research needs-to-be-fast crypto code so there is no need to get my speed fixes in in a hurry)

thomwiggers:
Actually, #25 is also on-hold, and #27 should probably come after that.

PR #27, opened 2018-06-20
Speed up binary operations by eliminating branches

thomwiggers:
I've benchmarked the binary operations and came to ~1250 ns per iteration.
Eliminating the branches, it's now down to ~750 ns.

### DIFF
--- a/benches/vob.rs
+++ b/benches/vob.rs
@@ -2,8 +2,10 @@
 
 #[macro_use]
 extern crate vob;
+extern crate rand;
 extern crate test;
 
+use rand::Rng;
 use test::Bencher;
 use vob::*;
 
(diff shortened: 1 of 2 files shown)


ltratt:
Are those numbers the wrong way around? i.e. it was 750 before and now it's 1250?

thomwiggers:
no, I left out the time unit.

ltratt:
OK, yes, that makes sense. I can also replicate the effect locally, so overall I like this.

I suggest that this might better be done in a single commit (though if you prefer separating them, I'm OK with that too). I think we can drop the `// Don't use short-circuiting || as it'll introduce a branch.` comments, as that's explained in the commit message.

ltratt:
If we document every design / performance decision as a comment, we'll be overloaded :) I think correctness comments are vital, but I don't think anyone's going to think "I'll change this to `||`" without using the nice benchmarks you've provided (at which point they'll change their mind). So I think it would be good to remove these comments.

thomwiggers:
Ok, done it.

ltratt:
Looks good. If you squash, I can merge.

thomwiggers:
done

PR #28, opened 2018-06-30
Make a Vob 1.3.2 release

ltratt:


Cargo.toml
```diff
@@ -2,11 +2,9 @@
 name = "vob"
 description = "Vector of Bits with Vec-like API and usize backing storage"
 repository = "https://github.com/softdevteam/vob/"
-version = "1.3.1"
+version = "1.3.2"
 authors = ["Laurence Tratt <laurie@tratt.net>"]
 readme = "README.md"
-# This should be "Apache-2.0 OR MIT OR UPL-1.0" but crates.io doesn't know about
-# UPL-1.0.
 license = "Apache-2.0 OR MIT"
```
thomwiggers: If you're removing that comment, don't you want to also add `OR UPL-1.0`?

ltratt: At least when I last checked, crates.io didn't support UPL-1.0 as a license, but I shouldn't have deleted the comment, which is a reminder to check at some point in the future. So I'm going to revert that commit (which should probably be squashed away before merging).

ltratt:
Oops, I completely missed this message! Squashed (and sorry for the delay!).

PR #30, opened 2018-08-08
Benchmark tweaks

ltratt:
First homogenise the benchmarks a bit (https://github.com/softdevteam/vob/commit/033201ffb4cfe49274a3ae5da0ed350ff654856c) and then add a benchmark for iter_set_bits (https://github.com/softdevteam/vob/commit/1caa09a2a0001c13d9aa66a4ce6c094ba4726e89) (which I vaguely feel we could probably make even faster one day, if we have a use case).


thomwiggers:
Needs to be updated against master though.

ltratt:
Oops! Rebased.

PR #39, opened 2019-02-04
added to_bytes() function and tests

farhanpatwary:
I've added a to_bytes() function which returns the contents of a Vob in byte form.
For example:
```let mut x = Vob::from_elem(3, true);```
calling ```to_bytes()``` on ```x``` would return ```[0b10100000]```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -481,6 +481,43 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
         }
     }
 
+    ///Returns contents of Vob as Vec of bytes.
+    ///
+    ///For a Vob of length n, the first bit will be the 0th element,
+    ///the last bit would be the (n-1)th element.
+    ///
+    /// # Examples
+    /// ```
+    ///use vob::Vob;
+    ///let mut x = Vob::from_elem(5, false);
+    ///x.set(2, true);
+    ///x.set(4, true);
+    ///assert_eq!(x.to_bytes(), [0b00101000]);
+    ///```
+    pub fn to_bytes(&self) -> Vec<u8> {
+        fn bit<T: Debug + One + PrimInt + Zero>(vob: &Vob<T>, byte: usize, bit: usize) -> u8 {
+            let offset = byte * 8 + bit;
+            if offset >= vob.len {
+                0
+            } else {
+                (vob[offset] as u8) << (7 - bit)
+            }
+        }
+
+        let len = self.len / 8 +
+                  if self.len % 8 == 0 { 0 } else { 1 };
+        (0..len).map(|i|
+            bit(self, i, 0) |
+            bit(self, i, 1) |
+            bit(self, i, 2) |
+            bit(self, i, 3) |
+            bit(self, i, 4) |
+            bit(self, i, 5) |
+            bit(self, i, 6) |
+            bit(self, i, 7)
+        ).collect()
+    }
+
     /// Returns an iterator over the slice.
     ///
     /// # Examples
@@ -1660,4 +1697,20 @@ mod tests {
         v.push(true);
         assert_eq!(v.vec.len(), 1);
     }
+
+    #[test]
+    fn to_bytes() {
+        let mut x = Vob::from_elem(3, true);
+        x.set(1, false);
+        assert_eq!(x.to_bytes(), [0b10100000]);
+    }
+
+    #[test]
+    fn to_bytes_larger(){
+        let mut x = Vob::from_elem(9, false);
+        x.set(2, true);
+        x.set(8, true);
+        assert_eq!(x.to_bytes(), [0b00100000, 0b10000000]);
+    }
+
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -490,29 +490,23 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
     /// assert_eq!(x.to_bytes(), [0b00101000]);
     /// ```
     pub fn to_bytes(&self) -> Vec<u8> {
-        fn bit<T: Debug + One + PrimInt + Zero>(vob: &Vob<T>, byte: usize, bit: usize) -> u8 {
-            let offset = byte * 8 + bit;
-            if offset >= vob.len {
-                0
-            } else {
-                (vob[offset] as u8) << (7 - bit)
+        let mut v: Vec<u8> = Vec::new();
+        for j in 0..self.vec.len() {
+            let mut current_block = self.vec[j];
+            let bytes_per_block = bytes_per_block::<T>();
+            for i in (0..bytes_per_block).rev() {
+                let x = current_block
+                    & num_traits::cast::cast(
+                        255_u64.rotate_left(8 * (bytes_per_block - 1 - i) as u32),
+                    )
+                    .unwrap();
+                let y: u8 = num_traits::cast::cast(x >> 8 * (bytes_per_block - 1 - i)).unwrap();
+                if y != 0 {
+                    v.push(y.reverse_bits());
+                }
             }
         }
-
-        let len = self.len / 8 + if self.len % 8 == 0 { 0 } else { 1 };
-        let x = (0..len)
-            .map(|i| {
-                bit(self, i, 0)
-                    | bit(self, i, 1)
-                    | bit(self, i, 2)
-                    | bit(self, i, 3)
-                    | bit(self, i, 4)
-                    | bit(self, i, 5)
-                    | bit(self, i, 6)
-                    | bit(self, i, 7)
-            })
-            .collect();
-        return x;
+        return v;
     }
 
     /// Returns an iterator over the slice.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/lib.rs [hunk 1]
```diff
@@ -481,6 +481,43 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+                  if self.len % 8 == 0 { 0 } else { 1 };
+        (0..len).map(|i|
+            bit(self, i, 0) |
+            bit(self, i, 1) |
+            bit(self, i, 2) |
+            bit(self, i, 3) |
+            bit(self, i, 4) |
+            bit(self, i, 5) |
+            bit(self, i, 6) |
+            bit(self, i, 7)
```
thomwiggers: This can be done much more efficiently. It is possibly to use `reverse` on nightly.

src/lib.rs
```diff
@@ -1660,4 +1697,20 @@ mod tests {
...
+        assert_eq!(x.to_bytes(), [0b10100000]);
+    }
+
+    #[test]
+    fn to_bytes_larger(){
+        let mut x = Vob::from_elem(9, false);
+        x.set(2, true);
+        x.set(8, true);
+        assert_eq!(x.to_bytes(), [0b00100000, 0b10000000]);
+    }
```
thomwiggers: I would like to see a test that shows exactly 8 elements and a test for 0 elements.

thomwiggers: I would like to see a test that shows exactly 8 elements and a test for 0 elements.

src/lib.rs
```diff
@@ -481,6 +481,43 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
         }
     }
 
+    ///Returns contents of Vob as Vec of bytes.
```
ltratt: All other coments in this file are `/// XXX` i.e. a space after `///`. Also, the first sentence currently doesn't parse.

You also need to run `cargo fmt` over this in order to shut Travis up.

src/lib.rs
```diff
@@ -481,6 +481,43 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+    ///```
+    pub fn to_bytes(&self) -> Vec<u8> {
+        fn bit<T: Debug + One + PrimInt + Zero>(vob: &Vob<T>, byte: usize, bit: usize) -> u8 {
+            let offset = byte * 8 + bit;
+            if offset >= vob.len {
+                0
+            } else {
+                (vob[offset] as u8) << (7 - bit)
+            }
+        }
```
thomwiggers: This seems unnecessary, as most of this already happens in `get` (except for the shift).

src/lib.rs
thomwiggers: You could probably do `iter_set_bits` to grab a more efficient slice as well.

src/lib.rs
farhanpatwary: Yes

ltratt: It sounds like you're using an old version of rustfmt. Personally I use the nightly version.

farhanpatwary: Thanks, I used ```cargo +nightly install --force rustfmt-nightly``` and then ran ```cargo +nightly fmt``` and now it works fine

src/lib.rs
farhanpatwary: Hi, I've pushed changes that are ready to be reviewed

src/lib.rs
```diff
@@ -476,6 +476,52 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+    /// x.set(2, true);
+    /// x.set(4, true);
+    /// assert_eq!(x.to_bytes(), [0b00101000]);
+    /// ```
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut v: Vec<u8> = Vec::new();
+        for j in 0..self.vec.len() {
+            let current_block = self.vec[j];
+            let bytes_per_block = bytes_per_block::<T>();
+            for i in (0..bytes_per_block).rev() {
```
ltratt: Here's an idea: if we treat the block as an array of bytes we *might* be able to speed things up a bit. Here's a sketch of the idea https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=854448102030bc9658a7a069e6011a6b -- notice that we'd need to see benchmark numbers before and after to know if the extra complexity is worth it.

src/lib.rs
```diff
@@ -476,6 +476,52 @@ impl<T: Debug + PrimInt + One + Zero> Vob<T> {
...
+                let x = current_block
+                    & num_traits::cast::cast(
+                        255_u64.rotate_left(8 * (bytes_per_block - 1 - i) as u32),
+                    )
+                    .unwrap();
+                let y: u8 = num_traits::cast::cast(x >> 8 * (bytes_per_block - 1 - i)).unwrap();
+                if y != 0 {
```
ltratt: This condition should probably be moved inside the `#[cfg(not(reverse_bits))]` clause. It *probably* slows things down for the `#[cfg(reverse_bits)]` case but, again, we should benchmark this.

src/lib.rs
```diff
@@ -1655,4 +1701,43 @@ mod tests {
         v.push(true);
         assert_eq!(v.vec.len(), 1);
     }
+
+    #[test]
+    fn to_bytes_1_byte_a() {
```
ltratt: To  match the names of other tests, this should be something like `test_to_bytes_single` and this should be merged with `to_bytes_1_byte_b`.

src/lib.rs
```diff
@@ -1655,4 +1701,43 @@ mod tests {
...
+    }
+    #[test]
+    fn to_bytes_1_byte_b() {
+        let mut x = Vob::from_elem(8, true);
+        x.set(1, false);
+        assert_eq!(x.to_bytes(), [0b10111111]);
+    }
+
+    #[test]
+    fn to_bytes_empty() {
```
ltratt: This should be the first test of this lot, since it's the base case.

src/lib.rs
```diff
@@ -1655,4 +1701,43 @@ mod tests {
...
+        let x = Vob::new();
+        assert_eq!(x.to_bytes(), []);
+    }
+
+    #[test]
+    fn to_bytes_2_bytes() {
```
ltratt: I don't think this test is super necessary.

src/lib.rs
ltratt: Actually, `to_bytes_1_byte_b` is *exactly* the same test, so we should delete that one.

src/lib.rs
```diff
@@ -1655,4 +1701,43 @@ mod tests {
...
+        let mut x = Vob::from_elem(9, false);
+        x.set(2, true);
+        x.set(8, true);
+        assert_eq!(x.to_bytes(), [32, 128]);
+    }
+
+    #[test]
+    fn to_bytes_using_binary_a() {
```
ltratt: This test is another duplicate of `test_to_bytes_single`.

src/lib.rs
```diff
@@ -1655,4 +1701,43 @@ mod tests {
...
+        let x = Vob::from_elem(1, true);
+        assert_eq!(x.to_bytes(), [0b10000000]);
+    }
+
+    #[test]
+    fn to_bytes_using_binary_b() {
```
ltratt: Ditto.

farhanpatwary: Hi, I've made the majority of the changes.
I didn't try to change the implementation of to_bytes to use a bit array as you suggested.
For the tests, I removed the ones you mentioned and added a test to test to_bytes on a Vob with more than one word, it's a simple test because I wasn't sure on how to create a test that tests to_bytes on random Vobs. I made all the other changes you have mentioned though

src/lib.rs
```diff
@@ -1703,15 +1706,14 @@ mod tests {
...
+    fn to_bytes_single_byte() {
         let x = Vob::from_elem(8, true);
         assert_eq!(x.to_bytes(), [255]);
     }
     #[test]
-    fn to_bytes_1_byte_b() {
-        let mut x = Vob::from_elem(8, true);
-        x.set(1, false);
-        assert_eq!(x.to_bytes(), [0b10111111]);
+    fn to_bytes_mutliple_words() {
```
ltratt: Spelling: "multiple".

ltratt:
> I didn't try to change the implementation of to_bytes to use a bit array that you suggested.

I don't know what this is referring to. Any clues?

src/lib.rs
farhanpatwary: whoops my bad, I've fixed that now

farhanpatwary:
> > I didn't try to change the implementation of to_bytes to use a bit array that you suggested.
> 
> I don't know what this is referring to. Any clues?

Here's an idea: if we treat the block as an array of bytes we might be able to speed things up a bit. Here's a sketch of the idea https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=854448102030bc9658a7a069e6011a6b -- notice that we'd need to see benchmark numbers before and after to know if the extra complexity is worth it.

ltratt:
That wasn't suggesting a bit array: it. Anyway, we need to try that before the PR can be considered for merging. You'll need to benchmark before and after that change so we can see if it's worth it (my gut feeling is that it will be worth it, but I've been wrong about many such things before).

ltratt:
This has timed out.

PR #42, opened 2019-08-10
Use stabilized features

thomwiggers:
`Bounds::*` have been available for a long while now, `reverse_bits` is about to be stabilized


ltratt:
Ah, I hadn't noted the `Bounds` thing, but I had noticed that `reverse_bits` is about to be stabilised. Works for me!

PR #43, opened 2019-08-10
don't use `extern crate` anymore in examples

thomwiggers:
We're on Rust 2018 now.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -155,13 +155,10 @@ impl Vob<usize> {
     /// Create a Vob from a `u8` slice. The most significant bit of each byte comes first in the
     /// resulting Vob.
     ///
-    /// If you are running nightly, this method will use the new `reverse_bits` intrinsic.
-    ///
     /// # Examples
     ///
     /// ```
-    /// #[macro_use] extern crate vob;
-    /// use vob::Vob;
+    /// use vob::{Vob, vob};
     /// fn main() {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -159,13 +159,13 @@ impl Vob<usize> {
     /// Create a Vob from a `u8` slice. The most significant bit of each byte comes first in the
     /// resulting Vob.
     ///
-    /// If you are running nightly, this method will use the new `reverse_bits` intrinsic.
+    /// If you are running rustc 1.37 or later, this method will use the new `reverse_bits`
```
ltratt: I think we can drop this part of the comment entirely: the performance of the method isn't something we really want to commit to in the comment IMHO.

ltratt:
Somehow this one entirely passed me by -- sorry! One minor comment, then I think we can get this in.

ltratt:
Assuming that new commit does everything you want (it does what I want!), please squash.

thomwiggers:
If you approve, you can just use the "squash and merge" button, right?

ltratt:
Unfortunately doing it via GitHub means that we get mixed-up commit messages :( So it's better if you squash it yourself, sort out the commit message, force push, and then I merge. Sorry for the extra steps :/

thomwiggers:
I don't think that in this case it would've mattered much.

ltratt:
Thanks!

PR #44, opened 2019-08-10
Fix rand crate version specification

thomwiggers:
Because it was specified as `0.6`, it would try to use a not backwards-compatible version.


thomwiggers:
ok it was complaining super hard about all sorts of incompatibilities, but now it isn't anymore when I do rustup update..

PR #48, opened 2019-11-25
Clean up doctests

thomwiggers:
Remove unnecessary ``fn main`` wrappers from doctest examples.

Also cleans up remaining ``#[macro use] extern crate vob;``
imports that were missed in #43

I was playing around with ``cargo clippy`` and it started complaining
about this.


ltratt:
Simplifications are always good -- thanks!

PR #50, opened 2020-08-30
Reserving space in a Vec gives a minimum, not a specific, guarantee.

ltratt:
Vec::reserve states:

>  Reserves capacity for at least additional more elements to be inserted in the
>  given Vec<T>. The collection may reserve more space to avoid frequent
>  reallocations.

so our previous test that we got *exactly* the size we asked for is incorrect (and, on my local machine, actually fails!). Update the tests accordingly.

PR #51, opened 2020-08-30
Create and initialise the Vec in one go.

ltratt:
Suggested by Clippy: this probably makes no difference, but it might conceivably allow the compiler to generate more efficient code, and it's no worse to read than before.

Will almost certainly need https://github.com/softdevteam/vob/pull/50 merged first.


thomwiggers (approved):

thomwiggers:
Clippy is amazing.

ltratt:
bors try

ltratt:
@thomwiggers Looks like we don't even need to rebase these, as by merging with master, bors/buildbot gets access to the config it needs.

PR #53, opened 2021-01-18
Add no-mut get_storage function

thomwiggers:
We've added `get_storage_mut()` a while ago, but it makes sense to also expose the non-mut equivalent.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -872,10 +872,10 @@ impl<T> Vob<T> {
     /// # Assumptions:
     ///
     /// * `Vob` stores the bits in a little-endian order.
-    /// * The length can't change, or should be updated using `Vob::set_len()`.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -888,6 +888,20 @@ impl<T> Vob<T> {
         &mut self.vec
     }
 
+    /// Get a reference to the underlying data structure
+    ///
+    /// `Vob` stores the bits in little-endian order. Any beyond `self.len()` should be `0`.
```
ltratt, line 893: At the risk of sounding like a lawyer, in retrospect this should say "will" rather than "should". Similarly we should update the comment "Any bits past `self.len()` should be set to 0 in the last block." above to saying "must be set to 0".

ltratt:
Agreed, this makes sense.

ltratt:
Please squash.

Issue #56, opened 2021-08-16
vob! macro parameter order


kocsis1david:
You need to specify the size of the Vob differently than Vec:

```
values: vec![0; width * height],
mask: vob![width * height; false],
```

It seems that the parameters of `from_elem` is also mixed.

ltratt:
Good point. Let me chew for a bit as to whether the API breakage is worth fixing this or not.

PR #57, opened 2021-09-18
Reorder from_elem and vob! arguments.

ltratt:
This is in response to #56. The basic idea here is that we inherited `from_elem`'s argument order from `BitVec` but that is the opposite order to other things in rust's main source repository (not that any of them are public). Because of that the `vob![x; y]` macro took its arguments in opposite order to `vec!`. Although I don't feel strongly about this (mostly because it's a breaking API change), it does feel like it makes more sense to mirror `Vec` than `BitVec` here. Any thoughts @thomwiggers?

[FWIW, I "ported" grmtools to this interface to see how difficult/easy it is: it was only a few minutes work, and I use `from_elem` repeatedly extensively in that codebase. rustc's errors make pinpointing the necessary changes trivial.]

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1229,7 +1229,7 @@ fn blocks_required<T>(num_bits: usize) -> usize {
 /// v2.push(true);
 /// v2.push(false);
 /// assert_eq!(v1, v2);
-/// println!("{:?}", vob![10; true]);
+/// println!("{:?}", vob![true; 10]);
 /// ```
 macro_rules! vob {
     (@single $($x:tt)*) => (());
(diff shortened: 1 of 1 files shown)


thomwiggers:
I think this actually has tripped me up in the past, although I don't think I thought it through far enough to ever realise why that was. I don't think we should be afraid to make this change to the API, it's an easy fix for downstream consumers and that's why there's semver anyway.

thomwiggers (approved):

ltratt:
If you're happy, I'm happy!

thomwiggers:
oh I probably should do 

bors r+

PR #59, opened 2021-09-20
Prepare a 3.0.0 release

ltratt:
Note there is one code change here (https://github.com/softdevteam/vob/commit/8e8a7c1f44bb08a902ace5784886e65c01c25f79) though it *should* be a "simple" alpha-renaming.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,16 +16,19 @@
 * Clearly document the invariants users must maintain when using the `unsafe`
   parts of the API.
 
+
 # vob 2.0.5 (2020-09-02)
 
 * Use rustfmt stable.
 * Remove unnecessary ``fn main`` wrappers in doctest examples.
 * Clean up two remaining ``#[macro use] extern crate vob;`` lines from examples.
 
+
 # vob 2.0.4 (2019-11-21)
 
 * License as dual Apache-2.0/MIT (instead of a more complex, and little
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,54 +1,75 @@
+# vob 3.0.0 (2021-09-20)
+
+## Breaking changes
+
+* `Vob::from_elem(value: bool, len: usize)` now takes the value first and the
+  number of repetitions of that value second to mirror the most common way this
+  function is defined elsewhere.
```
thomwiggers, line 7: I think it would be better if this was slightly more explicit that we mean `from_elem` as it appears in e.g. `Vec`.

thomwiggers, line 7:
```suggestion
* `Vob::from_elem(value: bool, len: usize)` now takes the value first and the
  number of repetitions of that value second to mirror the most common way
  `from_elem` is defined elsewhere, such as in `Vec`.
```

thomwiggers (changes requested):

CHANGES.md
ltratt: Weirdly, `Vec` does not define `from_elem` -- even though other related types in the Rust repo do!

thomwiggers: oh derp. disregard this then.

thomwiggers (approved):
bors r+

PR #62, opened 2021-10-17
Bit or assign and friends

### LINKED ISSUE #61: Operations |=, &=, ^=
Do you want a [PR](https://github.com/eadf/vob/commit/87c879dae1e58fa6385359b190b3d1ab3a68ac8e) for the std::ops::Bit*Assign operations?

eadf:
Close #61

~Work in progress: We need to figure out the issues raised in #61.~ 
The PR is ready for review.

Note: I will squash this before release, so commit comments might get lost.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,7 @@ use std::{
     slice,
 };
 
-use num_traits::{One, PrimInt, Zero};
+use num_traits::{PrimInt, Zero};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
@@ -189,7 +189,7 @@ impl Vob<usize> {
     }
 }
 
-impl<T: Debug + PrimInt + One + Zero> Vob<T> {
+impl<T: Debug + PrimInt> Vob<T> {
     /// Constructs a new, empty Vob (with a user-defined backing storage type) with the given
     /// capacity.
     ///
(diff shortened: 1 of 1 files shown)


eadf:
Just noticed that `padded_xor()`, `padded_and()` and `padded_or()` conditionally needs a call to `other._mask_last_block();` (or a workaround that fixes that issue)
But I won't spend time fixing that until you confirm that this the way you want the Bit traits implemented.

ltratt:
> I just realized that all of the operations that has a Self:Output field can return a Result<Vob<T>,Err<_>>.

True... but as you point out `(a & b)?` is weird and I'm not sure what `Err` would contain either. I guess part of my thinking is that when you do something with `Vob`s of unequal length, it's going to cause a crash sooner or later -- and it's probably best to do so sooner rather than later. Note that I can be convinced otherwise if you (or maybe @thomwiggers) has a good argument for this!

src/lib.rs
```diff
@@ -1251,6 +1242,225 @@ macro_rules! vob {
...
+    /// ```
+    /// use vob::vob;
+    /// let v1 = vob![true, false, true, false];
+    /// let v2 = vob![true, true, false, false];
+    /// let v3 = &v1 | &v2;
+    /// assert_eq!(v3, vob![true, true, true, false]);
+    /// ```
+    #[inline(always)]
+    fn bitor(self, other: &Vob<T>) -> Vob<T> {
+        let mut rv = self.clone();
```
ltratt, line 1288: Since this consumes `self`, can't we do `let _ = self.or(other)` as in `bitor_assign`?

ltratt: Ah, this is for `&Vob<T>`. Ignore my comment.

ltratt:
All the trait implementations look correct to me... but I think we need tests for each :) We don't need to heavily test them, but at least one test per trait implementation will keep us honest in the future.

eadf:
Isn't the doc tests enough for that purpose? They are run at each `cargo test` as well as the normal tests. Don't get me wrong, - I would not object to writing more tests if you like

PR #64, opened 2022-04-30
Fix incorrect assert in test

thomwiggers:
This test should be `assert_eq!`. Currently it was using "true" as the format string argument.


thomwiggers:
I found this one when I was trying to upgrade to edition 2021 for fun.

ltratt:
Good spot!

bors r+

PR #70, opened 2023-11-11
Move to Criterion for benchmarking.

ltratt:
`cargo bench` is still a nightly feature, which makes it easy not to notice that changes have broken benchmarking (my desktop, for example, exclusively runs stable).

This commit moves us from Rust's in-built benchmarking system to Criterion. This increases the boiler-plate but a) it works on stable (so we can now also test in CI!) b) it provides better statistical output.

### DIFF
--- a/.buildbot.sh
+++ b/.buildbot.sh
@@ -14,6 +14,7 @@ cargo test
 cargo test --features "unsafe_internals"
 cargo test --release
 cargo test --release --features "unsafe_internals"
+cargo bench
 
 cargo fmt --all -- --check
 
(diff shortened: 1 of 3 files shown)


eadf (commented):
I usually encapsulate the thing I'm most interested in benchmarking in a `criterion::black_box()` block, not that i have any proof that it makes any difference.
 -It's just for good measure. 
`black_box()` makes sure that the compiler does not optimize away the anything important.

ltratt:
`black_box` is a fun one, in part because it's not guaranteed to work! I would tend to use it in benchmarks to stop constant propagation. I don't think that's a big problem here -- or, if it is, it was a problem before, and I haven't made it worse since I haven't removed any `black_box`es. So I'm inclined to keep this PR `black_box` free, but if someone then wanted to investigate whether we *should* have been using `black_box` and finds places where that causes a difference, that would make a fine follow-up PR.

thomwiggers (approved):

PR #71, opened 2023-11-12
Added a benchmark comparing Vob, Vec and HashSet.

eadf:
I've noticed a 25% difference in performance when removing the `black_box()` in this benchmark. While I understand the preference to exclude it, this significant performance change raises some questions. I'm open to removing it as per your preference, but I wanted to bring this observation to your attention for discussion.

### DIFF
--- a/benches/comparative.rs
+++ b/benches/comparative.rs
@@ -1,0 +1,81 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use rand::prelude::*;
+use std::collections::HashSet;
+
+const SIZE: usize = 10000;
+
+/// This benchmark compares the performance of HashSet, Vob, and Vec
+/// in a scenario where an algorithm needs to keep track of visited indices.
+#[cfg(test)]
+fn comparative(c: &mut Criterion) {
+    // Generate 10000 random (seeded) points
+    let mut rng: StdRng = SeedableRng::from_seed([42; 32]);
+    let mut indices: Vec<usize> = (0..SIZE).collect();
+
+    indices.shuffle(&mut rng);
+
+    c.bench_function("HashSet comparative", |b| {
+        b.iter(|| {
+            let mut hash_set = HashSet::new();
+            // set half of the set to "true"
+            for index in indices.iter().skip(SIZE / 2) {
+                hash_set.insert(index);
+            }
+
+            for index in indices.iter().take(SIZE / 2) {
+                assert!(
+                    !hash_set.contains(index),
+                    "index {} is supposed to be cleared",
+                    index
+                );
+            }
+            for index in indices.iter().skip(SIZE / 2) {
+                assert!(
+                    hash_set.contains(index),
+                    "index {} is supposed to be set",
+                    index
+                );
+            }
+        })
+    });
+
+    c.bench_function("Vob comparative", |b| {
+        b.iter(|| {
+            let mut vob: vob::Vob<u32> = vob::Vob::new_with_storage_type(SIZE);
+            vob.resize(SIZE, false);
+
+            // set half of the vob to true
+            for index in indices.iter().skip(SIZE / 2) {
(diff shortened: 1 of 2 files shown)


benches/comparative.rs
```diff
@@ -0,0 +1,66 @@
...
+
+const SIZE: usize = 10000;
+
+/// This benchmark compares the performance of HashSet, Vob, and Vec
+/// in a scenario where an algorithm needs to keep track of visited indices.
+#[cfg(test)]
+fn comparative(c: &mut Criterion) {
+    // Generate 10000 random (seeded) points
+    let mut rng: StdRng = SeedableRng::from_seed([42; 32]);
+    let mut indices = Vec::<usize>::new();
```
ltratt, line 13: I think this can be something like `let mut indices = 0..SIZE..collect()`?

eadf: Good catch, fixed. 
I'd like to rebase the PR before release though

ltratt:
Don't worry, we'll give you a squash+rebase opportunity! I'll also ask if @thomwiggers has any comments.

benches/comparative.rs
```diff
@@ -0,0 +1,64 @@
...
+    c.bench_function("HashSet comparative", |b| {
+        b.iter(|| {
+            let mut hash_set = HashSet::new();
+            // set half of the set to "true"
+            for index in indices.iter().skip(SIZE / 2) {
+                black_box(hash_set.insert(index));
+            }
+
+            for index in &indices {
+                black_box(hash_set.contains(index));
```
thomwiggers, line 26: this usage of `black_box` only makes sure contains is used, both `index` and `hash_set` are still subject to optimization. 

This seems fine in this example, but I just wanted to flag it.

https://doc.rust-lang.org/std/hint/fn.black_box.html

ltratt:
Thanks @thomwiggers !

@eadf Please squash + rebase + force push.

benches/comparative.rs
eadf: I think the objective is not to prevent optimization but to prevent the optimizer from removing the call altogether. 
If the compiler sees a side-effect free method call where the output is not used, it could elect not to do the call at all (that is at least my 5 cent understanding). 

Maybe we could accomplish the same thing by moving the black box outside the call?:

```rust
for index in &indices {
     let visited = hash_set.contains(index);
     black_box(!visited);
```

thomwiggers: Yeah, I think the main thing is that the compiler might still e.g. unroll this loop or const-eval `hash_set.contains()`.

ltratt: At some point we can become unduly paranoid. But I do see that you both have made a good point. Perhaps we need two levels:

```rust
for index in blackbox(&indices) {
  hash_set.contains(blackbox(index));
}
```

The first `blackbox` (hopefully) prevents the compiler realising that the `for` loop's length can be statically known. The second stops it realising that the index is (by defintion) contained in the loop. *But* since the `contains` has no side-effect, technically it can optimise it away. So I'd probably actually force a side effect:


```rust
for index in blackbox(&indices) {
  if !hash_set.contains(blackbox(index)) {
    panic!("...");
  }
}
```

thomwiggers: The latter should probably be written using `assert!()` instead of an explicit `panic!()`, though.

eadf: The tests are setup so that only half of the values are "true", or else the Vob would get an unfair advantage by testing if the entire u32 is set. (not that i actually seen how that test is actually implemented).

We could re-arrange the setup so we could cross check if the bit should be true or not, though.

thomwiggers: Either way I think we're over-engineering this. @eadf could probably quickly check if additional `black_box` around index makes any difference and then put this to bed?

ltratt:
Thanks both!

PR #73, opened 2025-03-15
Add optional bincode dependency

ratmice:
bincode 2.0 has an alternative to the serde `Serialize`/`Deserialize` traits.  This patch adds an optional dependency and derives that trait.

* bincode serde is [optional](https://docs.rs/bincode/latest/bincode/#serde) 
* [Decode trait](https://docs.rs/bincode/latest/bincode/de/trait.Decode.html)
* [Encode trait](https://docs.rs/bincode/latest/bincode/enc/trait.Encode.html)

Also silences a rustc lint that appears to have changed between 1.37.0 and now.
I'm not exactly certain the compiler version that change happened in, so I used the version that I was running, I would imagine the warning probably was added a lot earlier though.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ rustc_version = "0.4"
 [dependencies]
 num-traits = "0.2.14"
 serde = { version="1.0", features=["derive"], optional=true }
+bincode = { version = "2.0", features=["derive"], optional=true }
 
 [dev-dependencies]
 criterion = "0.5"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,8 @@ use std::{
     slice,
 };
 
+#[cfg(feature = "bincode")]
+use bincode::{Decode, Encode};
 use num_traits::{PrimInt, Zero};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
@@ -89,6 +91,7 @@ use serde::{Deserialize, Serialize};
 /// `Vob`'s [`set_all(false)`](struct.Vob.html#method.set_all) function.
 #[derive(Clone, Default)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "bincode", derive(Encode, Decode))]
 pub struct Vob<T = usize> {
     /// How many bits are stored in this Vob?
     len: usize,


build.rs
```diff
@@ -6,5 +6,8 @@ fn main() {
     // Set features depending on channel
     if version().unwrap() >= Version::parse("1.37.0").unwrap() {
         println!("cargo:rustc-cfg=reverse_bits");
+        if version().unwrap() >= Version::parse("1.85.0").unwrap() {
```
ltratt, line 9: I *think* but I might be wrong this can be:

```rust
if version > 1.85.0 { println!("...") }
else if version > 1.37.0 { println!("...") }
```

but I might be wrong!

ratmice: It at least works on 1.85 if you don't print the other.

Changed in [a8b2625](https://github.com/softdevteam/vob/pull/73/commits/a8b262562eb5ec1c2c91b07ca757eff6598c5f2d)

ltratt:
Please squash.

build.rs
ratmice: Oh wait hold up, I believe this other line is *changing* the cfg value,
e.g. it is equivalent to RUSTFLAGS="--cfg reverse_bits".

while `rustc-check-cfg` is just silencing a warning that you are using a `#[cfg(foo)]` that rustc doesn't know about (whether `cfg(reverse_bits)` is set or not.

So I think this changes the default of whether that `--cfg` is passed to depend on the compiler.

I don't think that is what we want?

ltratt: My memory on this is fuzzy, but I think `reverse_bits` was stabilised at some point? Maybe it doesn't even need mentioning at all for recent Rust versions?

ratmice: I *think* what is going on is that `reverse_bits` is a *vob* config option that turns on the fast path.
If we remove the emitting of that line we revert back to the slow path on 1.85 and later.

So I think we want both `println` on 1.85 and just the original `1.37` and later

ratmice: Or maybe consider removing the slow path code, if it has been long enough?

ltratt: I think there are two aspects:

1. `reverse_bits` turns on a fast path by using the `reverse_bits` function in std which was stabilised in 1.37.
2. The way we tell Cargo to turn on that fast path has changed.

I tend to agree that supporting < 1.37 (Aug 2019) no longer feels worthwhile. I will raise a PR to remove the slow path.

ratmice: Great, I'll just remove the `build.rs` changes from this patch and squash, leaving it to that PR.

ratmice:
Squashed with the build.rs change removed.

PR #74, opened 2025-03-16
No longer try to support ancient Rust versions.

ltratt:
We've kept around support for pre-Rust-1.37 for long enough: it's unlikely that anyone using a pre-Aug-2019 rustc will be brave/foolish enough to update Vob, or any other crate, any more!

[Spotted in #73, because the Cargo syntax has changed, so if we wanted to keep this hack going forwards, we'd need to have another conditional for newer rustc versions. That doesn't really seem worth the complexity.]


thomwiggers (approved):
sgtm

thomwiggers (changes requested):
`Cargo.toml` should also be updated:

```toml
[build-dependencies]
rustc_version = "0.4"
```

Cargo.toml
```diff
@@ -10,7 +10,7 @@ categories = ["data-structures"]
 edition = "2018"
 
 [build-dependencies]
-rustc_version = "0.4"
+rustc_version = "1.37"
```
thomwiggers, line 13: I don't think this is necessary at all anymore?

thomwiggers: Since it's only for `build.rs` and that got deleted.

ltratt: Good point. Fixed in https://github.com/softdevteam/vob/pull/74/commits/44c967c1755c0abdb19956b09a9191c2d1e62c6c.

thomwiggers:
Seems the merge queue doesn't care about resolved conversations and approval, by the way

ltratt:
Merge queues make my head hurt!

ltratt:
Ah, `cargo deny` needs a tweak before anything in this repo will merge. Fixed.

thomwiggers:
#76 duplicates that but okay

ltratt:
Dumb question: should this lead to a major release, or are we comfortable with this being a minor release? I *think* minor releases fairly frequently break "this works on old rustc" but I'm unsure whether that's technically "in the spirit of the semver rules" or not.

thomwiggers:
I think projects that are more explicit about MSRV do do semver breaks when they change the minimum rustc required; I don't think this project ever made such claims (Certainly, we don't seem to have the metadata).  https://github.com/rust-random/rand/commit/c1f865f10ffecc8a211e98ac7e8c5014d3949f45

ratmice:
I think that it is generally considered maintainer judgement, in that there is nothing holding anyone to a specific compiler version.

Like you can specify updating the minimum supported rust version is a breaking change as a guarantee that you commit to as maintainer, but unless you do that there is nothing to hold you to it.

ltratt:
Thanks both. I think a minor release seems most appropriate then.

Issue #78, opened 2025-06-06
Question: How do you feel about adding get/set_unchecked()?


eadf:
To squeeze out the last bits of performance we could add unchecked access methods, similarly to how [Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get_unchecked) is implemented:

```rust
///.....
/// # Safety
/// The caller must ensure that `index < self.len()`.
/// Accessing an index >= self.len() is undefined behavior.
pub unsafe fn get_unchecked(&self, index: usize) -> bool {
    let blk = *self.vec.get_unchecked(block_offset::<T>(index));
    blk & (T::one() << (index % bits_per_block::<T>())) != T::zero()
}

///.....
/// # Safety  
/// The caller must ensure that `index < self.len()`.
/// Accessing an index >= self.len() is undefined behavior.
pub unsafe fn set_unchecked(&mut self, index: usize, value: bool) -> bool {
    let msk = T::one() << (index % bits_per_block::<T>());
    let off = block_offset::<T>(index);
    let old_v = *self.vec.get_unchecked(off);
    let new_v = if value { old_v | msk } else { old_v & !msk };
    if new_v != old_v {
        *self.vec.get_unchecked_mut(off) = new_v;
        true
    } else {
        false
    }
}
```

ltratt:
Please raise a PR. The only comment I've got is that the "For a safe alternative see [`get`]." sentences should be on the same line as the preceding lines. Newlines are expensive ;)

PR #79, opened 2025-06-06
Added unchecked_get() and unchecked_set()

eadf:
Updated the "comparative" benchmarks.
Added inline on bits_per_block() and bytes_per_block()

Related: #78


ltratt:
Of course, I realised something: we need some tests. Just checking that `get` and `unchecked_get` return the same results for a `random_vob` should do.

ltratt:
Thanks!

ltratt:
Please force push a `cargo fmt` update.

PR #81, opened 2025-07-07
Double ended iteration for `iter_set_bits`.

ltratt:
The main commit here is https://github.com/softdevteam/vob/commit/f9372b482770e08e8b7029cb4943367fa60f386f.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1101,6 +1101,10 @@ impl<'a, T: Debug + PrimInt> IntoIterator for &'a Vob<T> {
     }
 }
 
+// In all of the iterators below, we try to optimise what we expect to be the common cases (all
+// bits set or all bits unset) with the general case (random bit patterns). We assume that
+// `leading_zeros` and `trailing_zeros` have efficient implementations on most CPUs.
+
 #[derive(Clone)]
 pub struct IterSetBits<'a, T: 'a> {
     vob: &'a Vob<T>,
@@ -1113,24 +1117,15 @@ impl<T: Debug + PrimInt> Iterator for IterSetBits<'_, T> {
     fn next(&mut self) -> Option<usize> {
         debug_assert!(self.range.end <= self.vob.len);
         if let Some(i) = self.range.next() {
-            // This is a long, but fairly fast, way of finding out what the next set bit is. The
-            // basic problem is that we have no idea where the next set bit is -- but different
-            // patterns of set bits are most efficiently handled by different code paths. This code
-            // is thus a compromise: we prioritise two special cases (all bits set; all bits unset)
-            // for efficiency, and try and make the other possible cases reasonably fast.
             let mut b = block_offset::<T>(i);
             let mut v = self.vob.vec[b];
-            // If all bits are set, we don't need to do any complicated checks.
             if v == T::max_value() {
                 return Some(i);
             }
-            // At this point we've got a block which might or might not have some bits set. We now
-            // fall back to the general case.
             let mut i_off = i % bits_per_block::<T>();
             loop {
                 let tz = (v >> i_off).trailing_zeros() as usize;
                 if tz < bits_per_block::<T>() {
-                    // There is a bit set after i_off in the block.
                     let bs = b * bits_per_block::<T>() + i_off + tz;
                     self.range.start = bs + 1;
                     if bs >= self.range.end {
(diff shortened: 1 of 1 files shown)


ltratt:
I don't think it would add expense. I don't know whether it's very useful, or not, though; but if it is useful, and it's safe for us to do so, I'm happy to see a PR for it.

ratmice:
> I'm not totally sure of `ExactSizeIterator`'s guarantees: note that we do change `self.range`, but we only ever make it smaller. If that is allowable

Is that just the normal size reduction by one corresponding to the consuming of an item from the iterator?
Because that is expected for `ExactSizeIterator`, (also I see that `Iter` implements it, but `IterSetBits` and `IterUnsetBits` do not).

also out of curiousity are the iterators fused?  (never return `Some` after returning `None`) it seemed like they might not be if that `loop` can fall-through in the `Bits` iterators.

1. if they are this is potentially another trait that could be implemented
2. Curiously reading through the `ExactSizeIterator` docs I'm uncertain the expected behavior for a non-fused iterator.

Since the docs kind of imply that the iterator should be fused

[ExactSizeIterator::len](https://doc.rust-lang.org/stable/std/iter/trait.ExactSizeIterator.html#method.len)
> The implementation ensures that the iterator will return exactly `len()` more times a `Some(T)` value, before returning `None`.

So I'm curious what a non-fused iterator should return, the number of `Some` items remaining until the next `None` I suppose. (As opposed to the number of `Some` items remaining in total which in some cases might be more useful for sparse collections?).

eadf:
> iter_set_bits has no idea how many times it will return Some

But if iter_set_bits is fused, then we can figure that out:

`size_hint()` just need to do something like this (pseudo code)
```
fn count_set_bits_in_range(&self, range: Range<usize>) -> usize {
    let start_word = range.start / 64;
    let end_word = range.end / 64;
    
    if start_word == end_word {
        // Range entirely within one word
        let mut word = self.words[start_word];
        let start_bit = range.start % 64;
        let end_bit = range.end % 64;
        
        // Shift out unwanted bits directly
        word <<= start_bit;           // Remove bits before start
        word >>= start_bit + (64 - end_bit);  // Remove bits after end
        return word.count_ones() as usize;
    }
    
    let mut count = 0;
    
    // First word: shift out bits before start_bit
    let start_bit = range.start % 64;
    let first_word = self.words[start_word] >> start_bit;
    count += first_word.count_ones() as usize;
    
    // Middle words (unchanged)
    for word_idx in (start_word + 1)..end_word {
        count += self.words[word_idx].count_ones() as usize;
    }
    
    // Last word: shift out bits after end_bit
    if end_word < self.words.len() {
        let end_bit = range.end % 64;
        let last_word = self.words[end_word] << (64 - end_bit);
        count += last_word.count_ones() as usize;
    }
    
    count
} 
```
The question is if we are willing to pay that O(number-of-words) cost.

ltratt:
I agree that it works, but I'm not sure there's a use case which is important enough to bear this (surprising) cost: I think most users would expect the cost to be O(1). [Another way of putting this is that the only way to know how far long the journey will be is to complete the journey then go back and tell everyone how long it will be :)]

ltratt:
Dumb question: how does `ExactSizeIterator` optimise `vob::iter_set_bits().collect()`.

Good point about `FusedIterator` I must admit that I didn't even know this exists! I'll raise a PR shortly.

eadf:
> how does ExactSizeIterator optimise vob::iter_set_bits().collect()

`.collect()` first creates a `Vec::with_capacity(vob.size_hint())` and then populates it from the iterator, so if the hint is large, the capacity is large.

eadf:
We could always implement size_hint() as:
```
  fn size_hint(&self) -> (usize, Option<usize>) {
           // Conservative estimate - maybe small heuristic or just (0, None)
           let conservative = (self.vob.len() / 100).max(64);
           (0, Some(conservative))
    }
```

But i just had an idea, why not have the best of both worlds?

```
 pub fn iter_set_bits<const EXACT: bool = false, R>(&self, range: R) -> IterSetBits<T>
    where
        R: RangeBounds<usize>,
    {
    ...
    
```

So normally we just use the current, fast, variant with the same syntax as before. 

If the user uses` EXACT=true` we calculate the exact `size_hint()`.

We could not implement `ExacSizeIterator` in this scenario, as the return type should remain the same in both cases.  
And the `vob.iter_set_bits::<true,Range<usize>>()` syntax is bit awkward, but only used when really needed.

ltratt:
Is there much use in this if we can't implement `ExactSizeIterator`?

PR #85, opened 2025-07-11
Make blocks_required() branchless.

eadf:
`bits_per_block()` and `bytes_per_block()` now compiles into `usize` literals: the ultimate inline.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1304,30 +1304,27 @@ impl<T: Debug + PrimInt> Iterator for StorageIter<'_, T> {
 
 #[inline(always)]
 /// How many bits are stored in each underlying storage block?
-fn bits_per_block<T>() -> usize {
+const fn bits_per_block<T>() -> usize {
     bytes_per_block::<T>() * 8
 }
 
 #[inline(always)]
 /// How many bytes are stored in each underlying storage block?
-fn bytes_per_block<T>() -> usize {
+const fn bytes_per_block<T>() -> usize {
     size_of::<T>()
 }
 
+#[inline(always)]
 /// Return the offset in the vector of the storage block storing the bit `off`.
-fn block_offset<T>(off: usize) -> usize {
+const fn block_offset<T>(off: usize) -> usize {
     off / bits_per_block::<T>()
 }
 
+#[inline(always)]
 /// Takes as input a number of bits requiring storage; returns an aligned number of blocks needed
 /// to store those bits.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1302,32 +1302,27 @@ impl<T: Debug + PrimInt> Iterator for StorageIter<'_, T> {
...
-fn blocks_required<T>(num_bits: usize) -> usize {
-    num_bits / bits_per_block::<T>()
-        + if num_bits % bits_per_block::<T>() == 0 {
-            0
-        } else {
-            1
-        }
+/// to store those bits. (Using the classic ceiling division formula: ⌈n / b⌉ = ⌊(n + b - 1) / b⌋)
+const fn blocks_required<T>(num_bits: usize) -> usize {
+    (num_bits + bits_per_block::<T>() - 1) / bits_per_block::<T>()
```
ltratt, line 1325: A problem with this (more on 32 than 64 bit) is that it can overflow, even with a valid (and on 32 bit, very plausible) size. Even worse, in release mode, it won't cause a panic -- something weird will happen, but I'm not quite sure what in this context!

ltratt: I'm allergic to `as`, but `usize::from` probably works and would be fine with me!

eadf: I'm not sure if replacing an `if` branch with a, potentially non-inlined, method call is any improvement:
```
fn blocks_required<T>(num_bits: usize) -> usize {
    num_bits / bits_per_block::<T>() + usize::from(num_bits % bits_per_block::<T>() != 0)
}
```

Just checked with [godbolt](https://godbolt.org/z/errWb4zfo): with "-C opt-level=3" both `as` and `usize::from()` compiles down to the same one-single-OP code.

ratmice: It definitely looks to me like the `from` impl just calls `as`, in an `#[inline(always)]` function.
https://doc.rust-lang.org/src/core/convert/num.rs.html#76-79

eadf: Yeah, `as` can be dicey. Specially on machines where `usize != u64`, but once wrapped inside std::from() it's the problem of the Rust maintainers. 
For this particular conversion: `bool -> usize`, the op is safe even if `usize` is an`u8`.

ltratt:
We're having some problems with github at our end, so this might take a little while to merge in. Don't worry, it will!

ltratt:
Please can you force push an update to this branch? Hopefully that will kick our CI back into life.

Issue #86, opened 2025-07-12
Question: count_set_bits() and count_unset_bits() PR


eadf:
I've been [playing around ](https://github.com/eadf/vob/blob/db946989811c4d6c62f90dc39cbaff9a7f951ed9/src/lib.rs#L598)trying to create a faster `.iter_unset_bits(..).count()` method (like we discussed before)
This is what i came up with, i have not actually benchmarked it but since it is processing entire words instead of bit-by-bit, it should be substantially faster. 

Is this something you want a PR for?

```rust
    /// Counts the number of set bits.
    /// Generates the same result as Vob::iterate_set_bits().count() only much faster.
    /// # Examples
    /// 
    /// use vob::Vob;
    /// let mut v = Vob::new();
    /// v.push(false);
    /// v.push(true);
    /// assert_eq!(v.iter_set_bits(..).count(), v.count_set_bits(..));
    /// 
    pub fn count_set_bits<R>(&self, range: R) -> usize
    where
        R: RangeBounds<usize>,
    {
        self._count_set_bits(self.process_range(range))
    }

    /// Counts the number of unset bits.
    /// Generates the same result as Vob::iterate_unset_bits().count() only much faster.
    /// # Examples
    ///  
    /// use vob::Vob;
    /// let mut v = Vob::new();
    /// v.push(false);
    /// v.push(true);
    /// assert_eq!(v.iter_unset_bits(..).count(), v.count_unset_bits(..));
    /// 
    pub fn count_unset_bits<R>(&self, range: R) -> usize
    where
        R: RangeBounds<usize>,
    {
        let range = self.process_range(range);
        (range.end - range.start) - self._count_set_bits(range)
    }

    /// Counts the number of set bits.
    /// This method assumes the range is processed with process_range()
    fn _count_set_bits(&self, range: Range<usize>) -> usize {
        // Early return for empty ranges
        if range.start >= range.end {
            return 0;
        }
        let start_word = block_offset::<T>(range.start);
        if start_word >= self.len {
            return 0;
        }
        // this -1 is safe since we already tested for range.start & range.end equality
        let end_word = blocks_required::<T>(range.end) - 1;

        if start_word == end_word {
            // Range entirely within one word
            let word = self.vec[start_word];
            let start_bit = range.start % bits_per_block::<T>();
            let end_bit = range.end % bits_per_block::<T>();

            // Remove bits before start_bit and bits after end_bit
            return count_ones_usize(if end_bit == 0 {
[...]

PR #87, opened 2025-07-13
Optimized .count() for the iterators.

eadf:
Added some modest optimizations for `.count()` of all the iterators
## Results
With a benchmark of 100_000 random bits, on my M1 Mac:
* `Vob<usize>::iter().count()` took 31_472 ps, now it takes 314 ps (100 times faster)
* `Vob<usize>::iter_set_bits(..).count()` took 178_860 ns, now it takes 198 ns (900 times faster)
* `Vob::iter_storage().count()` has been updated too (now it just does math instead of counting).

## Changes
Added a private `count_set_bits()` method that counts set bits in O(number of words) 
Fixed the deprecated `rand::Rng` methods in the benchmark code
Added tests and benchmarks for the new `iter*.count()` functionality

## Issues that needs resolve:
✓  ~~### move `count_set_bits()`~~ (delay to next semver update)
We could move `count_set_bits()` into the iterator if `IterSetBits` and `IterUnsetBits` were merged into one `BitsIterator`. This could be done without any runtime penalties, and we'd skip some duplicated boilerplate: 

`pub struct BitsIterator<'a, T: 'a, const BIT_STATE: bool> {...}`

We could do this is stages, for now we could simply copy&paste like this (to reduce risk of new bugs):
```
fn next(&mut self) -> Option<usize> {
    if BIT_STATE {
      // old IterSetBits::next() code
    } else {
      // old IterUnsetBits::next() code 
    }
}
```

### Remove benchmark
Once the gain is confirmed, the new `count_bits()` benchmark can be removed because it simply tests `range.count()`, and that's like benchmarking std::Range.

### Possibly add a new helper method
for `n % bits_per_block::<T>()`, just like `block_offset()` does `n / bits_per_block::<T>()`

✓ ~~### Tidy up the` .filter(|_|true)`~~
I use `iterator.filter(|_| true).count()` to trigger the old `.count()` behavior in the tests, maybe there are better ways to do that.

✓ ~~### `count_ones_usize()`~~
`count_ones_usize()` could safely use `unwrap_unchecked()` (or it could even safely use `as`)

(link #86)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,7 @@
 
 use std::{
     cmp::{min, PartialEq},
+    convert::TryFrom,
     fmt::{self, Debug},
     hash::{Hash, Hasher},
     iter::{DoubleEndedIterator, FromIterator, FusedIterator},
@@ -585,6 +586,58 @@ impl<T: Debug + PrimInt> Vob<T> {
         }
     }
 
+    /// Counts the number of set bits.
+    /// This method assumes the range is processed with process_range()
+    fn count_set_bits(&self, range: Range<usize>) -> usize {
+        // Early return for empty ranges
+        if range.is_empty() {
+            return 0;
+        }
+        let start_off = block_offset::<T>(range.start);
+
+        // this -1 arithmetic is safe since we already tested for range.start & range.end equality
+        let end_off = blocks_required::<T>(range.end) - 1;
+
+        if start_off == end_off {
+            // Range entirely within one word
+            let b = self.vec[start_off];
+            let start_bit = range.start % bits_per_block::<T>();
+            let end_bit = range.end % bits_per_block::<T>();
+
+            // Remove bits before start_bit and bits after end_bit
+            let count = if end_bit == 0 {
+                // end_bit = 0 means we want everything from start_bit to end of word
+                // After the right shift, we have what we want
+                b >> start_bit
+            } else {
+                // We want bits from start_bit to end_bit
+                // After right shift, we need to remove the high bits
+                (b >> start_bit) << (start_bit + bits_per_block::<T>() - end_bit)
+            }
+            .count_ones();
+            return usize::try_from(count).unwrap();
+        }
+
+        // First word: shift out bits before start_bit
+        let start_bit = range.start % bits_per_block::<T>();
+        let mut count = usize::try_from((self.vec[start_off] >> start_bit).count_ones()).unwrap();
+
+        // Middle words
+        for word_idx in (start_off + 1)..end_off {
+            count += usize::try_from(self.vec[word_idx].count_ones()).unwrap();
+        }
+
+        // Last word: shift out bits after end_bit
+        let end_bit = range.end % bits_per_block::<T>();
+        let count_ones = if end_bit == 0 {
+            // end_bit = 0 means we want to count the entire end_off word
+            self.vec[end_off].count_ones()
+        } else {
+            (self.vec[end_off] << (bits_per_block::<T>() - end_bit)).count_ones()
+        };
+        count + usize::try_from(count_ones).unwrap()
+    }
+
     /// Returns an iterator which efficiently produces the index of each unset bit in the specified
     /// range. Assuming appropriate support from your CPU, this is much more efficient than
     /// checking each bit individually.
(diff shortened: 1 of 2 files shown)


ltratt:
Could you update the benchmarks so that they all use the same units please? My pea brain misreads comparisons when the units change!

src/lib.rs
```diff
@@ -585,6 +585,57 @@ impl<T: Debug + PrimInt> Vob<T> {
         }
     }
 
+    /// Counts the number of set bits.
+    /// This method assumes the range is processed with process_range()
+    fn count_set_bits(&self, range: Range<usize>) -> usize {
+        // Early return for empty ranges
+        if range.start >= range.end {
```
ltratt, line 592: Is such a range valid? I suspect we could, and probably should, `panic` in such a case. It would be good to know what the std library tends to do in such cases.

src/lib.rs
```diff
@@ -585,6 +585,57 @@ impl<T: Debug + PrimInt> Vob<T> {
...
+            return 0;
+        }
+        let start_word = block_offset::<T>(range.start);
```
ltratt, line 595: In vob we use `off` not `word`. So this would be `start_off` (etc).

src/lib.rs
```diff
@@ -585,6 +585,57 @@ impl<T: Debug + PrimInt> Vob<T> {
...
+        if start_word >= self.len {
+            return 0;
+        }
+        // this -1 is safe since we already tested for range.start & range.end equality
+        let end_word = blocks_required::<T>(range.end) - 1;
+
+        if start_word == end_word {
+            // Range entirely within one word
+            let word = self.vec[start_word];
```
ltratt, line 604: Vob calls this `b`.

src/lib.rs
```diff
@@ -1327,6 +1396,16 @@ fn blocks_required<T>(num_bits: usize) -> usize {
     num_bits / bits_per_block::<T>() + usize::from(num_bits % bits_per_block::<T>() != 0)
 }
 
+/// Convenience function that calls T::count_ones() and converts the result to usize
+/// (The conversion is always safe even if T is u128 and usize is u16)
+#[inline(always)]
+fn count_ones_usize<T: PrimInt>(value: T) -> usize {
```
ltratt, line 1402: I'd prefer to call the `count_ones` std lib method directly rather than use this wrapper. The `try_from` is mildly ugly in source code, but it'll be optimised away, so it's not worth worrying about.

ltratt:
I like the direction of this! My comments are mostly shallow and making sure this code feels in keeping with the rest of Vob's source, but the general idea clearly works as-is.

> We could move count_set_bits() into the iterator if IterSetBits and IterUnsetBits were merged into one BitsIterator

This would be a semver breakage. It might be worth doing *after* the next release, but it would be nice to get the optimisation out to people first before we think about breaking the API.

>I use iterator.filter(|_| true).count() to trigger the old .count() behavior in the tests, maybe there are better ways to do that.

I don't have a better idea!

> count_ones_usize() could safely use unwrap_unchecked() (or it could even safely use as)

rustc will almost certainly optimise the unwrap away entirely on the platforms we know and care about (certainly, in all similar cases I've seen, and checked, it has done so, though obviously one can not 100% guarantee this), so the `try_from` is good, because if someone ever tries this on a platform with different integer constraints it will break in the right way.

src/lib.rs
ratmice: I was certain it wasn't valid, but turns out it is just an alias for `range.is_empty()`, and `std` seems to happily construct one

https://doc.rust-lang.org/src/core/ops/range.rs.html#147-149
```
    pub fn is_empty(&self) -> bool {

        !(self.start < self.end)

    }
```
https://doc.rust-lang.org/std/ops/struct.Range.html#method.is_empty
> Returns true if the range contains no items.

eadf: if nothing else, it's a catch for bad user input. iter_set_bits(234234..25232344) will make process_range() return (len(),len()) (if the ranges are out of bounds)
If we don't want to allow bad input like that, we'd need to update process_range() with a Result return type

ltratt:
> But it is unlikely that anybody stores an IterSetBits as just that type, instead of a impl Iterator

I quite often store iterator types, so I know one user who'll be affected ;)

ltratt:
One minor comment then I think we're ready to merge.

eadf:
>One minor comment then I think we're ready to merge.

And the other issues? i.e remove `count_bits()` bench and helper method for `n % bits_per_block::<T>()`?

eadf:
Ok, I'll remove the benches i mentioned earlier and won't not add a modulus helper method.

ltratt:
Thanks!

PR #89, opened 2025-08-14
Prepare a 3.0.6 release.

ltratt:


CHANGES.md
```diff
@@ -1,3 +1,12 @@
+# vob 3.0.6 (2025-08-14)
+
+* Transparently optimise `iter_{set|unset}_bits(...).count()` to use platform
+  "count bits" directly, rather than forcing the full iterator to execute.
+  Depending on your machine, this can lead to savings of 2-3 orders of magnitude.
+
```
eadf, line 6: I don't know if it worth mentioning, but we updated the 'normal' interator.count() too.
It now does math instead of exhausting the iterator (`range.count()`)

ltratt: I did think briefly think about adding that in, but I came down on the side of "don't distract readers from the major things that will benefit them". I think the `iter_bits.count()` optimisation is something that a *lot* of users will really benefit from.

PR #90, opened 2026-07-16
Since bincode is unmaintained use wincode instead.

ratmice:
This is part of a series of patches to various crates moving from the bincode crate to the wincode crate.
I believe the reverse order of dependencies is:

1. packedvec
2. sparsevec
3. vob
4. grmtools

I'm currently leaving how we want to deal with semver up in the air. Since these do change trait impls around,
It may be that as we release these we need to fiddle with the version numbers in `Cargo.toml` too.


ratmice:
Force pushed a rustfmt.

PR #91, opened 2026-07-16
Various Clippy fixes.

ltratt:
I'm not super keen on `BITS as usize`, but it is clearer than `size_of<...> * 8`, and the alternative is
`usize::try_from(iX::BITS).unwrap()` which is objectionable in other ways.


ltratt:
@ratmice Hopefully this one will go through now.

